PR #8, opened 2024-06-19
Update read me file

kiranmurali93:
Corrected the copy config file


deepakkumarnd:
@kiranmurali93 In my machine renaming the env example to .env or even env is not working I still have to do a `source env` to get it pick the environment variables. Have you updated the database.yml by any chance ? Otherwise rails wouldn't pick the env file by default.

deepakkumarnd:
yea you can do that

deepakkumarnd (approved):

PR #14, opened 2024-06-24
Implement home page

lallenfrancisl:


Gemfile
lallenfrancisl, line 1: I had an error because this was missing

PR #29, opened 2024-09-17
manager view with the analytical rep(chart) is created 

sadrisya:
click on the team from the side bar will show the manager dash board


app/views/shared/components/_fixed_side_bar.html.erb
```diff
@@ -24,7 +24,7 @@
           <% end %>
         </li>
         <li class="side-bar-item">
-          <%= link_to "#", class: "side-bar-item-inner" do %>
+          <%= link_to manager_path, class: "side-bar-item-inner" do %>
```
deepakkumarnd, line 27: can you skip this change for now ? I will update this separately. We need to make this change for mobile sidebar too.

config/routes.rb
```diff
@@ -81,6 +81,8 @@
   # for now dashboard path and root path are same.
   get "error_401" => "pages#unauthorized"
   get "dashboard" => "dashboards#index"
+  get "manager" => "dashboards#index_manager"
```
deepakkumarnd, line 84: skip this route, I will add the necessary changes to make this work automatically

app/views/dashboards/_index_manager_dashboard_desktop.html.erb
```diff
@@ -0,0 +1,72 @@
+<%= desktop_view do %>
```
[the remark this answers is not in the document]
sadrisya: <%= desktop_view do %> is replaced with <%= render "shared/components/desktop_layout" do %>

app/views/shared/components/_fixed_side_bar.html.erb
sadrisya: reverted

config/routes.rb
sadrisya: removed the path

deepakkumarnd (approved):

PR #127, opened 2024-10-21
Teams mobile view tab change controller is modified

### LINKED ISSUE #85: Simplify the team page in mobile view.
The mobile view in team page has a tab switching part. User can click on user or team tab and switch between the user/team tab.

team_tab_change_controller is used at the top of the page. This should be moved to the section where we need to switch the tab.

The code here can be structured to contain only two sections (or divs) containing a user tab or a team tab. Using targets we can either hide/show the tab based on what the user want to view.

sadrisya:
tab-change controller is modified ,controller is given to the specific component which  contains the tabs .

Fixes #85


app/javascript/controllers/teams_tab_change_controller.js
```diff
@@ -1,34 +1,23 @@
...
-    const selectedContent = this.teamContentTargets.find(target => target.dataset.tab === tab);
-    if (selectedContent) {
-      selectedContent.classList.remove("hidden");
-    }
+
+    this.setActiveTab(selectedTab);
   }
+
   setActiveTab(activeTab) {
     const tabs = this.element.querySelectorAll("a[data-tab]");
```
deepakkumarnd, line 18: Can we avoid querySelector here and rely only on targets

app/views/teams/_show_mobile.html.erb
```diff
@@ -54,12 +53,12 @@
         <% end %>
       <% end %>
     </div>
-    <div id="user" data-teams-tab-change-target="teamContent" class="" data-tab="user">
+    <div id="user" data-teams-tab-change-target="teamContent" class="pt-6" data-tab="user" data-controller="user-show">
```
deepakkumarnd, line 56: Do we need user-show controller ?

app/views/teams/_show_mobile.html.erb
```diff
@@ -54,12 +53,12 @@
...
       <div class="my-4 flex hidden justify-start" data-target="user-show.allUsersSection">
         <span class="text-xs font-semibold text-primary">All users</span>
       </div>
       <div class="my-4 flex justify-end hidden" data-user-show-target="showAllUsers">
-        <a href="#" data-action="click->user-show#toggleAllUsers" class="hidden flex items-center justify-center text-base font-normal">
+        <a href="#" data-action="click->user-show#toggleAllUsers" class="flex items-center justify-center text-base font-normal">
```
deepakkumarnd, line 61: This is intentionally kept as hidden because the feature is not implemented yet

app/views/teams/_show_mobile.html.erb
```diff
@@ -29,8 +28,8 @@
         <%= render 'shared/components/button_default_medium', label: 'View full dashboard', icon_name: 'icon-arrow-right' %>
       </div>
     <% end %>
-
-    <div class="my-6 flex justify-start gap-6">
+   <div data-controller="teams-tab-change">
+    <div class="mt-2 mb-6 flex justify-start gap-6">
       <a href="javascript:void(0);" data-action="click->teams-tab-change#changeTab" data-tab="user" class="text-base font-normal active-border">User</a>
```
deepakkumarnd, line 33: Can use use target name here to relate this link to the actual tab ?

app/views/teams/_show_mobile.html.erb
deepakkumarnd: Can use use target name here to relate this link to the actual tab ?

app/javascript/controllers/teams_tab_change_controller.js
```diff
@@ -1,34 +1,23 @@
 import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
-  static targets = ["teamContent", "allUsersSection", "showAllLink"];
+  static targets = ["teamContent"];
```
deepakkumarnd, line 4: we probably should name both user tab and team tab as separate or just name it as a generic name like "tabContent".

app/views/teams/_show_mobile.html.erb
deepakkumarnd: There are so many identifiers here id, target and data-tab, why do we need all these ?

deepakkumarnd (changes requested):

app/javascript/controllers/teams_tab_change_controller.js
sadrisya: Query selector Removed and   used " const selectedTab = event.currentTarget.dataset.tab;"

app/views/teams/_show_mobile.html.erb
sadrisya: user-show controller removed as it is not using.also IDs are removed ,and targets used to get tab link

app/views/teams/_show_mobile.html.erb
sadrisya: target name used here

app/views/teams/_show_mobile.html.erb
sadrisya: user-show controller is removed, can be used later if needed

app/javascript/controllers/teams_tab_change_controller.js
sadrisya: teamContent has changed to tabContent

deepakkumarnd (approved):

PR #197, opened 2024-11-05
Team list implemented as a div changed to list(ul), with vertical scrollbar

### LINKED ISSUE #156: Convert team listing to a list with vertical scrollbar
Team list is implemented as a div, it has to be made into a list (ul) with vertical scrollbar

sadrisya:
Fixes #156

Screenshot 2024-11-05 at 2 34 39 PM (2)

### DIFF
--- a/app/views/teams/_sub_teams.html.erb
+++ b/app/views/teams/_sub_teams.html.erb
@@ -19,6 +19,7 @@
     <% end %>
   </div>
 <% end %>
+
 <ul id="team-list-desktop" class="flex flex-col gap-4 mt-6 overflow-y-auto max-h-[240px]">
   <% team.sub_teams.each do |sub_team| %>
     <%= render "teams/team_desktop", team: sub_team %>
--- a/app/views/teams/_sub_teams.html.erb
+++ b/app/views/teams/_sub_teams.html.erb
@@ -19,9 +19,8 @@
     <% end %>
   </div>
 <% end %>
-
-<div id="team-list-desktop" class="flex flex-col gap-4 pt-6">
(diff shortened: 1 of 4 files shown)


app/views/teams/_sub_teams.html.erb
```diff
@@ -19,9 +19,8 @@
     <% end %>
   </div>
 <% end %>
-
```
deepakkumarnd, line 22: Do not remove these blank lines. They are kept for idnentifying the blocks

app/views/teams/_sub_teams.html.erb
```diff
@@ -19,9 +19,8 @@
...
-<div id="team-list-desktop" class="flex flex-col gap-4 pt-6">
+<ul class="team-list flex flex-col gap-4 mt-6 overflow-y-auto max-h-[240px]">
```
deepakkumarnd, line 22: what is the use of team-list class ?, have you tested the code ? If you remove the "id" then adding the new item won't reflect in the list.

deepakkumarnd (changes requested):

app/views/teams/_sub_teams.html.erb
sadrisya: It is a typo, even when I tested couldn't find any issue ,so missed it

app/views/teams/_sub_teams.html.erb
sadrisya: this is happening on auto format while saving

sadrisya: Added the Blank line

app/views/teams/_sub_teams.html.erb
sadrisya: ID is replaced

PR #374, opened 2024-12-04
A button dissabled class is added ,wich can be used to disable the button

### LINKED ISSUE #314: Add a disabled class for link and button
Add a disbled styling for links and buttons to easily identify the dead links and incactive buttons in the UI 

adding a class 'inactive' to a button or link shoudl make it non clickable and on hover it should show the disabled pointer style.

sadrisya:
Added a disabled class which will disable  link, button ,div also  form submit, text area etc 
Fixes #314


app/assets/stylesheets/course_progress.css
```diff
@@ -72,3 +72,8 @@ details[open] .icon-up-arrow-green {
 
 }
 
+.btn-disabled {
```
deepakkumarnd, line 75: This is not specific to course. It should not be added to this file. Background color should also be changed

sadrisya: moved to new css file

PR #463, opened 2024-12-27
Enabled bulled gem in test env

### LINKED ISSUE #240: Enable bullet gem in test env
Note: Enabling the bullet gem in the test environment will cause current tests to fail due to n+1 queries, which need to be fixed.

paulsojan:
Fixes #240


paulsojan:
@deepakkumarnd Please review

PR #489, opened 2025-01-08
New colors according to branding is added

sadrisya:
New color code added


app/assets/stylesheets/application.css
```diff
@@ -15,13 +15,12 @@
  */
 
 * {
-    font-family: 'Montserrat', sans-serif;
-    padding: 0;
-    margin: 0;
-    box-sizing: border-box;
-    color: #333333;
-    letter-spacing: 0.025em;
+  font-family: "poppins";
```
deepakkumarnd, line 18: Can we do these in steps ? First go for changing colors, replace the colors with the new color code mapping. Do not add any new color variables.

PR #668, opened 2025-02-10
View all users web and mobile pages created

### LINKED ISSUE #669: Create a "View All Users" Page Accessible from the "More" Button in the Teams Page
We need to create a new "View All Users" page that can be accessed via the "More" button on the Teams page. This page should display a complete list of users with relevant details.

sadrisya:
Added "View All Users" to the  "More" options on the Teams page to navigate to the new "View All Users" page.
Created a new view to list all users with relevant details.

Fixes #669
Screenshot 2025-02-10 at 2 17 25 PM

Screenshot 2025-02-10 at 2 13 24 PM
Screenshot 2025-02-10 at 2 13 37 PM


app/controllers/all_users_controller.rb
```diff
@@ -0,0 +1,5 @@
+class AllUsersController < ApplicationController
+  def index
```
deepakkumarnd, line 2: this should be moved to teams_controller, the action name can be all_users.

sadrisya: Moved to teams controller

PR #686, opened 2025-02-13
Integrated fast2sms

paulsojan:
Fixes #683

PR #712, opened 2025-02-19
ensure uniqueness of non null mobile numbers

deepakkumarnd:
Fixes #686


deepakkumarnd:
discarded because the migration change is not necessary

PR #763, opened 2025-03-13
Implemented impersonation

### LINKED ISSUE #739: Shadow account.
### Requirement

Admins should be able to login as a shadow user (Support user) to any of the learning partner. This is needed because admins might need to perform some account management on behalf of the learning partner. Think this as an account manager doing some help such as inviting users on behalf of a learning partner or setting up an initial team etc. 

1. The shadow user may have a role called "account_manager" or "supervisor" or even "support". 
2. There shall be only one shadow user per learning partner(Do not impose this by database constraint but by application code).
3. The shadow user should not be listed in the user listing for other owners or managers.
4. The shadow user should be added to the top level team. 
5. The shadow user should have the same access privileges as that of "owner".

paulsojan:
Fixes #739 

Demo: https://paulsojan.neetorecord.com/watch/630ef5f03743e2296e5a


paulsojan:
@deepakkumarnd Please review

app/controllers/teams_controller.rb
```diff
@@ -12,7 +12,7 @@ def new
   def show
     authorize @team
     @learning_partner = current_user.learning_partner
-    @members = @team.members
+    @members = @team.members.where.not(role: "support")
```
deepakkumarnd, line 15: I think it is safe to keep this condition in the members function itself.

config/credentials/test.yml.enc
deepakkumarnd, line 1: Any change in this file ?

PR #773, opened 2025-03-19
Not found Text appearing error solution

### LINKED ISSUE #513: Not found text is appearing at the bottom of the course page
Screenshot 2025-01-13 at 1 30 08 PM
WhatsApp Image 2025-01-13 at 1 33 16 PM

sadrisya:
"Not found" text appears incorrectly at the bottom of the page due to Turbo requests receiving a 404 Not Found response.
Fixes #513

**Summary**

- not_found is triggered when an invalid route, missing record, or explicit render status: :not_found is called.
- A Turbo request will trigger not_found.()
- Rails renders the full layout before appending "Not found" as raw text.
- The browser appends this text outside <body> because it's not part of the structured HTML.
- Turbo requests encountering a 404 error do not reload the full page but append the response as-is.

Screenshot 2025-03-17 at 9 47 37 AM
Screenshot 2025-03-17 at 9 50 38 AM

**Root Cause**

- Turbo fetches a non-HTML response (text/vnd.turbo-stream.html), which appends "Not found" outside the structured         DOM.
- The Rails controller returns plain text instead of an HTML-rendered error page, causing the misplaced text issue.
- Stimulus controller makes a Turbo request immediately on page load, fetching a non-existent resource (/notifications/count) and triggering a 404.

**Solution**
Delayed the Turbo request using setTimeout to prevent an immediate 404.


app/javascript/controllers/notifications_controller.js
```diff
@@ -7,17 +7,19 @@ export default class extends Controller {
...
+          .then((response) => response.text())
+          .then((html) => Turbo.renderStreamMessage(html))
+          .catch((error) => Logger.error(error));
+      }, interval * 1000);
+
+      // clear the interval after 24 hours
+      setTimeout(() => {
+        clearInterval(checkNotificationInterval);
+      }, 86400 * 1000);
+    }, 180000);
```
deepakkumarnd, line 22: there is a bit of lack of clarity here and too many nesting insdie the setTimeout function. Can you refactor this to make it bit more readable. 

You can make the blocks as function and call the function inside the timeouts. Add comments to understand why the 180000 is addeded  etc.

deepakkumarnd (changes requested):

app/javascript/controllers/notifications_controller.js
sadrisya: Done the changes

PR #806, opened 2025-04-04
component for banner is created

### LINKED ISSUE #793: Create banner
Banner can be used in multiple cases, such as showing impersonation, payment plan expired etc.

sadrisya:
Fixes #793
Screenshot 2025-04-10 at 10 30 53 AM
Screenshot 2025-04-10 at 10 31 05 AM


deepakkumarnd:
This needs a bit of research for example the notificaiton bar has to be named notification_bar this is not a banner. 

Screenshot 2025-04-05 at 3 05 48 PM
Take a look at this as an example. This is something that we want to achive.

A tiny notification_bar with a background color, (our theme highlight color will do), black text on the center is what we need.

PR #841, opened 2025-05-13
Input field replaced with input components

### LINKED ISSUE #840: Use input components in the place of all input fields

sadrisya:
Fixes #840

### DIFF
--- a/app/helpers/ui_helper.rb
+++ b/app/helpers/ui_helper.rb
@@ -33,14 +33,17 @@ module UiHelper
     ordered.compact.join.html_safe # rubocop:disable Rails/OutputSafety
   end
 
-  def input_field(form: nil, field_name: nil, label: nil, placeholder: 'Enter text', width: 'w-56', left_icon: nil,
-                  right_icon: nil, type: 'text_field', options: [], value: 'nil')
+  def input_field(form: nil, field_name: nil, text_field_name: nil, radio_field_name: nil, label: nil,
+                  placeholder: 'Enter text', width: 'w-56', left_icon: nil, right_icon: nil, type: 'text',
+                  options: [], value: 'nil', rows: '5', option: nil)
     partial_path = "ui/inputs/#{type}"
     partial_path = 'ui/inputs/text_field' unless lookup_context.exists?(partial_path, [], true)
 
     render partial: partial_path, locals: {
       form:,
       field_name:,
+      text_field_name: text_field_name || field_name,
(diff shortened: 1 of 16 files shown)


app/views/devise/sessions/_new_desktop.html.erb
```diff
@@ -9,22 +9,15 @@
           <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
             <div class="flex flex-col gap-6">
               <%= render "devise/shared/error_messages", resource: resource %>
-              <div class="input-group">
-                <%= f.label :email, 'Email', class: "label-text" %>
-                <p class="mt-4">
-                  <%= f.email_field :email, class: "input-text", autofocus: true, autocomplete: "email" %>
-                </p>
+              <div class="input-group pt-2">
+                <%= input_field(form:f, field_name: :email, label: "Email", placeholder: "Enter your email",type: "email",width:"w-full") %>
```
deepakkumarnd, line 13: Code formatting

app/helpers/ui_helper.rb
```diff
@@ -33,22 +33,31 @@ def label_with_icon(icon, label_tag, position)
     ordered.compact.join.html_safe # rubocop:disable Rails/OutputSafety
   end
 
-  def input_field(form: nil, field_name: nil, label: nil, placeholder: 'Enter text', width: 'w-56', left_icon: nil,
-                  right_icon: nil, type: 'text_field', options: [], value: 'nil')
+  def input_field(form: nil, field_name: nil, text_field_name: nil, radio_field_name: nil, label: nil,
```
deepakkumarnd, line 36: why do we need field_name, text_field_name and radio_field_name ? input_field shouldn't abstract all sort of inputs ? For each input components we need to have separate components. Merging everything to a single component makes it very complicated.

app/helpers/ui_helper.rb
```diff
@@ -33,22 +33,31 @@ def label_with_icon(icon, label_tag, position)
...
+                  placeholder: 'Enter text', width: 'w-56', height: nil, left_icon: nil, right_icon: nil, type: 'text',
+                  options: [], value: nil, rows: '5', option: nil, html_options: {})
```
deepakkumarnd, line 38: there are options, option these namings makes the code really bad, it is not very easy to understand the thought process here. html_options is a hash here, other's are array and nil, this is very inconsistent.

deepakkumarnd (changes requested):

Issue #983, opened 2025-08-22
Add search functionality in user listing


deepakkumarnd:
When the manager goes to a team page he should be able to search users by name

PR #991, opened 2025-08-27
983 search users

deepakkumarnd:
Issue #983


app/controllers/user_searches_controller.rb
paulsojan, line 1: Maybe we can move all searches related to the controller to a different folder. Because we already have a controller named `SearchesController` to search for courses.

paulsojan (dismissed):

paulsojan (approved):

PR #1126, opened 2025-10-16
Dropdown component created

sadrisya:
Fixes : #1110


app/views/view_components/dropdown_component/_select_box.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if dropdown.form.present? %>
+  <%= dropdown.form.select dropdown.name,
+      options_for_select([["Select an option", ""]] + dropdown.options, dropdown.value),
```
dr-r3d: Can we option for custom "Select an option" label passes to dropdown_component. Also add flag to remove this in case when it is not required to display this

sadrisya: moved select options merge logic to helper to make placeholder optional and customizable

dr-r3d (approved):

PR #1130, opened 2025-10-17
Input text component fixes

dr-r3d:
fixes border radius in mobile 
fixes text component height change on focus by using ring css property to highlight instead of border


dr-r3d:
these fixed are moved to input_mobile_component PR due to their dependency on mobile inputs

Issue #1065, opened 2025-09-23
Genrating quiz from trascripts


deepakkumarnd:
Use transcripts and generate/regenerate quizz and answers.

PR #1165, opened 2025-11-13
1065 quiz generation

dr-r3d:
#1065

### DIFF
--- a/app/services/quizzes/generation_service.rb
+++ b/app/services/quizzes/generation_service.rb
@@ -44,14 +44,14 @@ module Quizzes
     end
 
     def summarized_transcripts
-      module_transcripts = course_module.lessons.includes(local_contents: :transcripts).collect do |lesson|
-        content = lesson.local_contents.in_english.first
-        {
+      course_module.lessons.each_with_object([]) do |lesson, accumulated|
+        next if lesson.transcript_text.blank?
+
+        accumulated << {
           title: lesson.title,
-          transcripts: (content && content.transcripts.pluck(:text).join(' ')) || ''
+          transcripts: lesson.transcript_text
         }
       end
-      module_transcripts.select { |lt| lt[:transcripts].present? }
     end
   end
 end
--- a/app/models/lesson.rb
+++ b/app/models/lesson.rb
@@ -29,6 +29,14 @@ class Lesson < ApplicationRecord
     service.diminished_rating(rating, last_rated_at)
   end
 
+  def transcripts
+    local_contents.in_english.first&.transcripts
+  end
+
+  def transcript_text
+    transcripts.present? ? transcripts.pluck(:text).join(' ') : ''
+  end
+
   private
 
   def unique_local_content_lang


lib/integrations/llm/gemini.rb
```diff
@@ -28,6 +32,11 @@ def generate_transcript(audio_file)
         ask(TRANSCRIPTION_PROMPT, file_path: audio_file, response_type: 'TranscriptSchema')
       end
 
+      def generate_quiz(count, transcripts)
+        prompt = "#{QUIZ_GENERATION_PROMPT.gsub('#COUNT', count.to_s)}\n```#{transcripts.to_json}```"
+        ask(prompt, response_type: 'QuizSchema')
+      end
```
paulsojan: We should also be able to generate quizzes with OpenAI, too. So create an implementation there, too.

app/controllers/quizzes_controller.rb
```diff
@@ -45,6 +45,42 @@ def update
...
+        Rails.logger.error("Invalid quiz data: expected 4 options, got #{options.count}")
+        next
+      end
+      quiz = @course_module.quizzes.new(
+        question: quiz_data['question'],
+        option_a: options[0],
+        option_b: options[1],
+        option_c: options[2],
+        option_d: options[3],
+        answer: quiz_data['answer_text']
```
paulsojan: The answer should be an option, not the entire answer.

Image

Like this

Image

app/services/quizzes/generation_service.rb
```diff
@@ -0,0 +1,55 @@
...
+    def initialize(course_module)
+      @course_module = course_module
+      @transcripts = summarized_transcripts
+    end
+
+    def generate_via_ai
+      return [] if transcripts.empty?
+
+      result = Integrations::Llm::Api.llm_instance(provider: :gemini).generate_quiz(MAX_QUIZ_QUESTIONS, transcripts)
+      return result.data['quizzes'] if result.ok?
```
paulsojan:
```suggestion
      return Result.ok(result.data['quizzes']) if result.ok?
```

app/controllers/quizzes_controller.rb
```diff
@@ -45,6 +45,42 @@ def update
...
+      )
+      if quiz.save
+        saved_count += 1
+        service.update_quiz_ordering!(@course_module, quiz, :create)
+      else
+        Rails.logger.error("Failed to save generated quiz: #{quiz.errors.full_messages.join(', ')}")
```
paulsojan: We need to alert the developer if quiz saving fails. In such cases, logging won't be the best option, instead, show a flash message or log to Sentry.

app/views/course_modules/_generate_quiz_button.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if @quiz_generation_service.generate? %>
```
paulsojan: Don't use `@` directly in partial, accept it via locals

lib/integrations/llm/gemini.rb
dr-r3d: Admin cant select LLM model. It requires code update. So can add implementation when we need to switch to openAI. Right now gemini implentation is sufficient

paulsojan:
@dr-r3d Please check https://paulsojan.neetorecord.com/watch/ded9b9fb08d96cb7d0e4

dr-r3d:
@paulsojan, you are testing previously generated quiz. I have updated quiz to add only answer label letter instead of the full answer text. please regenerate quiz. this should fix what you decribed.

paulsojan (approved):

PR #1217, opened 2025-12-21
Admin's course listing page is created

### LINKED ISSUE #1221: The checkbox tick icon color is white by default, which makes it visible on certain background colors even when the checkbox is not checked.
Image

Image

sadrisya:
1.Admin's course listing page is created 
2.new chip set is added
3.Appears filter box when closing search tags is fixed
4.fixes  #1221   :- Hide checkbox tick icon when unchecked and center it when checked

Screenshot 2025-12-21 at 10 32 28 PM

Screenshot 2025-12-22 at 11 41 18 AM

Screenshot 2025-12-22 at 11 12 40 AM
Screenshot 2025-12-22 at 11 12 51 AM

### DIFF
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -89,74 +89,76 @@
     </div>
     <%= paginate(@enrolled_courses) if @type.present? %>
   <% end %>
-  <% if current_user.is_admin? %>
-    <div class="flex flex-col gap-5 md:gap-6">
-      <div class="flex flex-col gap-4 mt-3">
-        <% @available_courses.each do |course| %>
-          <%= link_to course_path(course) do %>
-            <%= long_course_card_component(course: course) %>
+  <% if available_courses?(@available_courses) %>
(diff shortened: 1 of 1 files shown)


app/views/courses/index.html.erb
```diff
@@ -76,7 +89,7 @@
     </div>
     <%= paginate(@enrolled_courses) if @type.present? %>
   <% end %>
-  <% if available_courses?(@available_courses) && @type != "enrolled" %>
+  <% if available_courses?(@available_courses) && @type != "enrolled" && !policy(:course).new? %>
```
deepakkumarnd: Why is this needed ?

!policy(:course).new?

sadrisya: This check is added because the new course card (long card) is intended only for the admin page, while the learners page should remain unchanged. Without this condition, the same section would also be listed on the admin page, which is not expected. Hence, !policy(:course).new? is required.

deepakkumarnd (approved):

PR #1281, opened 2026-03-18
Add Neo AI webhook for course sync

### LINKED ISSUE #1275: Create a webhook controller for neo-ai
This controller will be notified once a course is generated by Neo-AI.

paulsojan:
Fixes #1275 

Adds a Neo AI webhook endpoint for course synchronization. When Neo AI finishes generating a course, it notifies this webhook, which then enqueues a background job to sync the course data into Blackboard LMS.

Currently, the course data fetched from Neo-AI is not stored in Blackboard. This is because we have yet to finalize the course structure provided by Neo-AI. Once we finalize that structure, we will need to create new columns in the course and its associated records for storing additional data.


dr-r3d:
## Why we closed PR 1281 and opened PR 1375 instead of modifying it

  PR 1281 implemented a webhook — Neo AI calls our endpoint when course generation
  completes, and we sync the data automatically in the background. That's a
  **system-triggered, push-based** flow.

  What we actually need is different: the user reviews the generated course in
  Content Studio and explicitly clicks "Save Course." Only then does the data get
  persisted to BlackboardLMS. That's a **user-triggered, pull-based** flow.

  The save button is an explicit user action — the user expects immediate
  confirmation that it worked. A background job breaks that contract: the page
  appears to succeed while the actual save may still be pending or may have failed
  silently.

  Synchronous also fits because there's no long-running work to offload. By the
  time the user reaches the save button, the course data has already been fetched
  and rendered on the structure page. The save is just committing that data to the
  database — a fast, bounded operation with no reason to defer it.

  ## Why a new PR

  Modifying PR 1281 would have buried both of these decisions inside a review
  thread, leaving the git history saying "we added a webhook" when what actually
  shipped was a synchronous save triggered by the user. Opening a fresh PR makes
  the intent explicit from the start: the title, description, and diff all
  describe exactly what was built and why.

PR #1366, opened 2026-05-13
Fix N+1 query (local): eager load lessons on course modules

sadrisya:
## Problem
Bullet detected an N+1 query when opening a course — course_modules were 
loaded without eager loading lessons, causing a separate DB query per 
module when the view called module_lessons.

## Fix
- Course#modules_in_order — added .includes(:lessons) once in the model. 
  Every caller (web show, helper, view, course_json.rb) gets the fix automatically
- CoursesController#show — reverted to @course.modules_in_order, 
  removing the duplicated inline fix
- Api::V1::CoursesController#show — dropped the now-redundant 
  course_modules: { lessons: ... } includes and preloads local_contents 
  separately after modules_in_order runs, since that's the only extra 
  association the API needs

## Why this approach
Fixing it in the model ensures all callers benefit from the fix rather 
than patching one action in isolation. The API controller is updated to 
avoid unnecessary duplicate includes.

### DIFF
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -4,8 +4,9 @@ module Api
   module V1
     class CoursesController < ApiController
       def show
-        course = Course.includes(:tags, course_modules: { lessons: :local_contents })
-          .find(params[:id])
+        course = Course.includes(:tags).find(params[:id])
+        lessons = course.modules_in_order.flat_map(&:module_lessons)
+        ActiveRecord::Associations::Preloader.new(records: lessons, associations: :local_contents).call
         render json: course.to_json_data, status: :ok
       rescue ActiveRecord::RecordNotFound
         render json: { error: 'Course not found' }, status: :not_found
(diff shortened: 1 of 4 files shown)


dr-r3d:
1. Fix location — description says set_course, diff is in show

  set_course (line 223) is untouched. The diff modifies line 67, which is inside show. The PR description is wrong about where the fix lands.

  2. Duplication — the PR replicates modules_in_order inline

  Course#modules_in_order (course.rb:108):
  RecordOrdering.records_in_order(course_modules, course_modules_in_order)

  The PR's fix in show:
  RecordOrdering.records_in_order(
    @course.course_modules.includes(:lessons),  # same call, just adds includes
    @course.course_modules_in_order
  )

  It's the same call, with .includes(:lessons) added — duplicated in the controller instead of fixing the model.

  3. There are other callers that won't get the fix

  app/models/concerns/course_json.rb:10     modules_in_order.each do |course_module|
  app/helpers/courses_helper.rb:142         course.modules_in_order
  app/views/courses/_course_lessons.html.erb modules_in_order(course).each_with_index

  All of these call the unfixed model method. If any of them access course_module.lessons, they still trigger N+1. The fix in show is isolated to one action while the model method — which is the shared
  entry point — stays broken.

dr-r3d (approved):

PR #1373, opened 2026-05-19
Add optional checkbox to long_course_card_component

sadrisya:
Summary
Added a checkbox param to long_course_card_component (defaults to false, no breaking change to existing callers).

Accepts false (hidden), true (bare checkbox), or a Hash of check_box_tag options — :name, :value, :checked, plus any extra HTML attributes.

Checkbox renders absolute top-right on the card; click is stopped from bubbling so it doesn't trigger the parent link_to.

Wired up in programs/_course_list.html.erb for privileged managers (manager mode only) and in programs/add_courses.html.erb for the course selection form.

ui_manifest.md updated to document the new param and its accepted shapes.


engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,4 +1,13 @@
 <div class="group rounded-xl border border-line-colour relative md:max-h-[132px]">
+  <% if checkbox %>
+    <% checkbox_opts = checkbox.is_a?(Hash) ? checkbox.dup : {} %>
+    <%= check_box_tag checkbox_opts.delete(:name) || 'checkbox',
+          checkbox_opts.delete(:value),
+          checkbox_opts.delete(:checked) || false,
+          class: 'absolute top-4 right-4 w-5 h-5 z-10 rounded border border-slate-grey-light checked:bg-primary checked:border-primary focus:ring-0 cursor-pointer',
+          onclick: 'event.stopPropagation()',
+          **checkbox_opts %>
```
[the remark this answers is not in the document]
sadrisya: The fix is minimal: checkbox_opts.delete(:id) extracts the caller-provided id (removing it from the splat so it won't appear twice), and id: checkbox_id passes it explicitly. When no :id was provided, checkbox_id is nil — Rails treats id: nil as "no id attribute", preventing the auto-derived id="checkbox" duplicate.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,4 +1,15 @@
 <div class="group rounded-xl border border-line-colour relative md:max-h-[132px]">
+  <% if checkbox %>
+    <% checkbox_opts = checkbox.is_a?(Hash) ? checkbox.dup : {} %>
+    <% checkbox_id = checkbox_opts.delete(:id) %>
+    <%= check_box_tag checkbox_opts.delete(:name) || 'checkbox',
```
dr-r3d: **Bug: `checkbox: true` produces a broken input.**

When `checkbox` is truthy but not a Hash, `checkbox_opts` becomes `{}`, so this call reduces to:

```ruby
check_box_tag("checkbox", nil, false, ...)
```

That renders `<input type="checkbox" name="checkbox" value="">` — a useless element with a hardcoded name and blank value. The `|| "checkbox"` fallback masks the problem rather than surfacing it.

The `ui_manifest.md` and PR description both advertise `true` as a valid "bare checkbox" value, but nothing useful is produced. Either guard against non-Hash truthy values and raise, or remove `true` as a documented option.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,4 +1,15 @@
...
+          checkbox_opts.delete(:value),
+          checkbox_opts.delete(:checked) || false,
+          id: checkbox_id,
+          class: 'absolute top-4 right-4 w-5 h-5 z-10 rounded border border-slate-grey-light checked:bg-primary checked:border-primary focus:ring-0 cursor-pointer',
+          onclick: 'event.stopPropagation()',
```
dr-r3d: **Inline JS is inconsistent with Stimulus conventions.**

The rest of the codebase uses Stimulus `data-action` attributes for behaviour (e.g. `action: "programs#toggleDeleteButton"` on the checkbox options passed in). Using a raw `onclick` attribute here is inconsistent and harder to test or CSP-harden.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,4 +1,15 @@
...
+          **checkbox_opts %>
```
dr-r3d: **`class:` collision will raise at runtime on Ruby 3.3.**

`ui_manifest.md` advertises `class` as a valid Hash key ("class overrides"). If a caller passes `class:` in the checkbox Hash, it survives the `delete` calls and ends up in `**checkbox_opts`. Since `class:` is already an explicit keyword argument above it, Ruby 3.3 raises `ArgumentError: keyword argument repeated` at render time.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: stop_propagation_controller.js — a single-method controller that will be auto-registered by eagerLoadControllersFrom. Tiny and reusable anywhere in the app.
Template — the onclick: is gone. The <span> wrapper carries data-controller and data-action, so stop-propagation is entirely Stimulus-driven and CSP-safe. Positioning classes (absolute top-4 right-4 z-10) moved to the span; the input stays clean. Crucially, **checkbox_opts remains on the check_box_tag call untouched — no risk of data: hash conflicts since the Stimulus attributes live on the wrapper, not the input.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: Helper — guard raises ArgumentError immediately when checkbox is truthy but not a Hash, so the silent broken-input case is impossible.
Manifest — removed true as a documented value; noted the ArgumentError and added :id to the documented Hash keys.
Spec — one new example covering the raised error.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: :class is now deleted from checkbox_opts before the splat and joined onto the base class string, so a caller-supplied class: is appended rather than causing a ArgumentError: keyword argument repeated at render time.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
dr-r3d: stop_propagation_controller.js should reside in neo_components engine, not is host app

sadrisya: done

app/assets/config/manifest.js
```diff
@@ -4,3 +4,4 @@
 //= link_tree ../../../vendor/javascript .js
 //= link_tree ../builds
 //= link content_studio/application.css
+//= link_tree ../../../engines/neo_component/app/javascript .js
```
dr-r3d: No need for add individual javascript file. the engine has a proper registration mechanism already in place. engine.rb registers the engine's own config/importmap.rb with importmap-rails via the neo_components.importmap initializer.

sadrisya: done

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
```diff
@@ -0,0 +1,230 @@
+# frozen_string_literal: true
```
dr-r3d, line 1: **Spec location** — this file lives in `spec/helpers/` (host app) but the component it tests lives in `engines/neo_component/`. It should be at `engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb` so the engine can be tested in isolation without the host app. As-is, if `UiHelper` ever stops including the engine helpers, this spec silently disappears from the engine's own test run.

dr-r3d (commented):
**Spec location** — this file lives in `spec/helpers/` (host app) but the component it tests lives in `engines/neo_component/`. It should be at `engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb` so the engine can be tested in isolation without the host app. As-is, if `UiHelper` ever stops including the engine helpers, this spec silently disappears from the engine's own test run.

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
sadrisya: Good catch. Moved the spec to engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb and added engines/neo_component/spec/rails_helper.rb alongside it. The rails_helper still bootstraps via the host app for now (no dummy app yet), but the spec is now inside the engine so bundle exec rspec engines/neo_component picks it up correctly. Left a comment in rails_helper pointing toward the dummy app as the next step.

dr-r3d (approved):

PR #1377, opened 2026-05-29
Rework assign course page with deadline modal form

sadrisya:
Move the form out of course_select_component and into a new deadline modal so that duration selection, course assignment, and submission all happen in one place. The Assign Course button now opens the modal instead of submitting directly.

Key changes:
- Add deadline_modal_component with its own form (id="deadline-form"); Apply to All section sits outside the form as a pure UI helper that mirrors selections into the per-course dropdowns instantly on change
- Add deadline_modal Stimulus controller: handles apply-all spreading, custom date propagation, cancel/reset, and turbo:submit-end to close the modal on success
- Update course_select_controller to inject hidden course_ids[] inputs into the modal form when it opens, replacing the old checkbox-in-form pattern
- Add deadline_duration_options helper to CoursesHelper
- Extend modal_box_component with wrapper_html_options and close_action args so callers can attach Stimulus targets and custom close handlers

Screenshot 2026-05-31 at 9 51 21 PM
Screenshot 2026-05-31 at 9 51 32 PM
Screenshot 2026-05-31 at 9 51 40 PM
Screenshot 2026-05-31 at 9 51 52 PM
Screenshot 2026-05-31 at 9 52 02 PM

## Summary by CodeRabbit

* **New Features**
  * Added a deadline selection modal when assigning courses, enabling users to set predefined or custom deadline dates per course or apply the same deadline to all selected courses.
  * Integrated date picker for custom deadline date selection.
  * Course assignment workflow now uses a modal-based interface for deadline configuration.

### DIFF
--- a/app/controllers/course_assigns_controller.rb
+++ b/app/controllers/course_assigns_controller.rb
@@ -16,15 +16,16 @@ class CourseAssignsController < ApplicationController
   end
 
   def create
-    course_ids = (params[:course_ids] || []).reject(&:blank?)
+    permitted = params.permit(course_ids: [], durations: {}, custom_dates: {})
+    course_ids = (permitted[:course_ids] || []).reject(&:blank?)
 
     if course_ids.empty?
       flash.now[:error] = 'No courses selected'
       return render
     end
 
-    durations_hash = params.fetch(:durations, {})
-    custom_dates_hash = params.fetch(:custom_dates, {})
+    durations_hash = permitted.fetch(:durations, {})
+    custom_dates_hash = permitted.fetch(:custom_dates, {})
     courses = Course.find(course_ids)
 
     @courses_with_deadline = courses.map do |course|
--- a/app/controllers/course_assigns_controller.rb
+++ b/app/controllers/course_assigns_controller.rb
@@ -16,19 +16,22 @@ class CourseAssignsController < ApplicationController
   end
 
   def create
-    course_ids = (params[:course_ids] || []).filter { |id| !id.empty? }
-    deadlines = (params[:duration] || []).map { |d| to_deadline(d) }
-
-    @courses_with_deadline = []
+    course_ids = (params[:course_ids] || []).reject(&:blank?)
 
     if course_ids.empty?
       flash.now[:error] = 'No courses selected'
       return render
     end
 
+    durations_hash = params.fetch(:durations, {})
+    custom_dates_hash = params.fetch(:custom_dates, {})
     courses = Course.find(course_ids)
 
-    @courses_with_deadline = courses.zip(deadlines)
+    @courses_with_deadline = courses.map do |course|
+      id = course.id.to_s
+      [course, to_deadline(durations_hash[id], custom_dates_hash[id])]
+    end
+
     service = Courses::ManagementService.instance
 
     if @team_assign
@@ -59,20 +62,15 @@ class CourseAssignsController < ApplicationController
     end
   end
 
-  def to_deadline(duration)
+  def to_deadline(duration, custom_date = nil)
+    return DateTime.parse(custom_date) if duration == 'custom' && custom_date.present?
+
     case duration
-    when 'one_day'
-      DateTime.now + 1.day
-    when 'two_days'
-      DateTime.now + 2.days
-    when 'one_week'
-      DateTime.now + 1.week
-    when 'two_weeks'
-      DateTime.now + 2.weeks
-    when 'one_month'
-      DateTime.now + 1.month
-    else
-      nil
+    when 'one_day' then DateTime.now + 1.day
+    when 'two_days' then DateTime.now + 2.days
+    when 'one_week' then DateTime.now + 1.week
+    when 'two_weeks' then DateTime.now + 2.weeks
+    when 'one_month' then DateTime.now + 1.month
     end
   end
 end
--- a/app/controllers/course_assigns_controller.rb
+++ b/app/controllers/course_assigns_controller.rb
@@ -73,8 +73,8 @@ class CourseAssignsController < ApplicationController
       return nil if custom_date.blank?
 
       begin
-        DateTime.parse(custom_date)
-      rescue ArgumentError
+        Time.zone.parse(custom_date)
+      rescue ArgumentError, TypeError
         nil
       end
     else
--- a/app/controllers/course_assigns_controller.rb
+++ b/app/controllers/course_assigns_controller.rb
@@ -32,6 +32,11 @@ class CourseAssignsController < ApplicationController
       [course, to_deadline(durations_hash[id], custom_dates_hash[id])]
     end
 
+    if @courses_with_deadline.any? { |course, dl| durations_hash[course.id.to_s] == 'custom' && dl.nil? }
+      flash.now[:error] = 'One or more custom dates are invalid'
+      return render status: :unprocessable_entity
+    end
+
     service = Courses::ManagementService.instance
 
     if @team_assign
@@ -63,14 +68,22 @@ class CourseAssignsController < ApplicationController
   end
 
   def to_deadline(duration, custom_date = nil)
-    return DateTime.parse(custom_date) if duration == 'custom' && custom_date.present?
+    if duration == 'custom'
+      return nil if custom_date.blank?
 
-    case duration
-    when 'one_day' then DateTime.now + 1.day
-    when 'two_days' then DateTime.now + 2.days
-    when 'one_week' then DateTime.now + 1.week
-    when 'two_weeks' then DateTime.now + 2.weeks
-    when 'one_month' then DateTime.now + 1.month
+      begin
+        DateTime.parse(custom_date)
+      rescue ArgumentError
+        nil
+      end
+    else
+      case duration
+      when 'one_day' then Time.current + 1.day
+      when 'two_days' then Time.current + 2.days
+      when 'one_week' then Time.current + 1.week
+      when 'two_weeks' then Time.current + 2.weeks
+      when 'one_month' then Time.current + 1.month
+      end
     end
   end
 end
--- a/app/javascript/controllers/course_select_controller.js
+++ b/app/javascript/controllers/course_select_controller.js
@@ -2,6 +2,7 @@ import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
     static targets = ["assignButton", "assignButtonInner", "modal"];
+    static outlets = ["deadline-modal"];
 
     toggleAssignButton() {
         const anyChecked = this.element.querySelectorAll('input[name="course_ids[]"]:checked').length > 0;
@@ -40,7 +41,6 @@ export default class extends Controller {
             }));
         }
 
-        this.modalTarget.style.display = "";
-        this.modalTarget.querySelector("button:not([disabled]), input:not([disabled]), select:not([disabled])")?.focus();
+        this.deadlineModalOutlet.open();
     }
 }
--- a/app/javascript/controllers/course_select_controller.js
+++ b/app/javascript/controllers/course_select_controller.js
@@ -31,9 +31,13 @@ export default class extends Controller {
 
         const container = this.modalTarget.querySelector('[data-deadline-modal-target="courseIdInputs"]');
         if (container) {
-            container.innerHTML = checkedIds
-                .map(id => `<input type="hidden" name="course_ids[]" value="${id}">`)
-                .join('');
+            container.replaceChildren(...checkedIds.map(id => {
+                const input = document.createElement('input');
+                input.type = 'hidden';
+                input.name = 'course_ids[]';
+                input.setAttribute('value', id);
+                return input;
+            }));
         }
 
         this.modalTarget.style.display = "";
--- a/app/javascript/controllers/course_select_controller.js
+++ b/app/javascript/controllers/course_select_controller.js
@@ -1,7 +1,7 @@
 import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
-    static targets = ["assignButton", "assignButtonInner", "modal"];
+    static targets = ["assignButton", "assignButtonInner"];
     static outlets = ["deadline-modal"];
 
     toggleAssignButton() {
@@ -22,7 +22,7 @@ export default class extends Controller {
         const checkedIds = [...this.element.querySelectorAll('input[name="course_ids[]"]:checked')]
             .map(cb => cb.value);
 
-        this.modalTarget.querySelectorAll('[data-deadline-modal-target="courseRow"]').forEach(row => {
+        this.deadlineModalOutlet.element.querySelectorAll('[data-deadline-modal-target="courseRow"]').forEach(row => {
             if (checkedIds.includes(row.dataset.courseId)) {
                 row.classList.remove("hidden");
             } else {
@@ -30,7 +30,7 @@ export default class extends Controller {
             }
         });
 
-        const container = this.modalTarget.querySelector('[data-deadline-modal-target="courseIdInputs"]');
+        const container = this.deadlineModalOutlet.element.querySelector('[data-deadline-modal-target="courseIdInputs"]');
         if (container) {
             container.replaceChildren(...checkedIds.map(id => {
                 const input = document.createElement('input');
--- a/app/javascript/controllers/course_select_controller.js
+++ b/app/javascript/controllers/course_select_controller.js
@@ -3,10 +3,6 @@ import { Controller } from "@hotwired/stimulus";
 export default class extends Controller {
     static targets = ["assignButton", "assignButtonInner", "modal"];
 
-    connect() {
-        this.showDuration = this.element.dataset.showDuration === "true";
-    }
-
     toggleAssignButton() {
         const anyChecked = this.element.querySelectorAll('input[name="course_ids[]"]:checked').length > 0;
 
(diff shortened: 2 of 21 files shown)


app/controllers/course_assigns_controller.rb
```diff
@@ -59,20 +62,15 @@ def set_user_or_team
     end
   end
 
-  def to_deadline(duration)
+  def to_deadline(duration, custom_date = nil)
+    return DateTime.parse(custom_date) if duration == 'custom' && custom_date.present?
```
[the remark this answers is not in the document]
sadrisya: Done — `8805a867`. Here's the review reply:

---

> All three findings addressed:
>
> - `DateTime.parse` is now wrapped in `begin/rescue ArgumentError` — malformed input returns `nil` instead of raising
> - `DateTime.now` replaced with `Time.current` throughout `to_deadline` so deadlines respect the app's configured timezone
> - `create` now checks for any `custom` course that resolved to a `nil` deadline and responds 422 with a flash error before hitting the service — the exception can no longer bubble as a 500
>
> All existing specs still pass.

engines/neo_component/app/views/view_components/course_select/_course_select_component.html.erb
```diff
@@ -1,23 +1,23 @@
...
+      <%= button_component(text: t("button.cancel"), type: "outline", size: "sm") %>
+    <% end %>
+    <button type="button"
+            class="flex-none"
+            disabled
+            data-course-select-target="assignButton"
+            data-action="click->course-select#openModal">
+      <%= button_component(text: "Assign Course", size: "sm", disabled: true,
+            html_options: { data: { course_select_target: "assignButtonInner" } }) %>
+    </button>
```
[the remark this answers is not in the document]
sadrisya: Done

app/javascript/controllers/deadline_modal_controller.js
```diff
@@ -0,0 +1,98 @@
...
+            row.querySelector('[data-deadline-modal-target="customDateWrapper"]')
+               ?.classList.add("hidden");
+        });
+
+        this.close();
+    }
+
+    close() {
+        this.element.closest(".modal-wrapper").style.display = "none";
+    }
```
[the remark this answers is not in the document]
sadrisya: The fix stores the result of closest() and only sets style.display when the element exists, so a missing .modal-wrapper ancestor silently no-ops instead of throwing a TypeError.

app/javascript/controllers/course_select_controller.js
```diff
@@ -1,8 +1,45 @@
...
+        });
+
+        const container = this.modalTarget.querySelector('[data-deadline-modal-target="courseIdInputs"]');
+        if (container) {
+            container.innerHTML = checkedIds
+                .map(id => `<input type="hidden" name="course_ids[]" value="${id}">`)
+                .join('');
+        }
+
+        this.modalTarget.style.removeProperty("display");
```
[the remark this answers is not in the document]
sadrisya: Asymmetric open/close — changed removeProperty("display") to style.display = "" so open and close are a matching pair
Focus on open — modal now focuses the first non-disabled button/input/select when it opens, so keyboard users land inside the modal immediately
Escape key — added a keydown listener that calls cancel() when the modal is visible
ARIA (aria-hidden, aria-modal, role="dialog") — skipped, the product does not currently support screen reader accessibility
Focus trap — skipped, out of scope for a minimal fix

app/controllers/course_assigns_controller.rb
```diff
@@ -16,19 +16,27 @@ def new
...
-    @courses_with_deadline = []
+    course_ids = (params[:course_ids] || []).reject(&:blank?)
 
     if course_ids.empty?
       flash.now[:error] = 'No courses selected'
       return render
     end
 
+    durations_hash = params.fetch(:durations, {})
+    custom_dates_hash = params.fetch(:custom_dates, {})
```
dr-r3d: `params.fetch(:durations, {})` and `params.fetch(:custom_dates, {})` bypass Rails strong parameters. In production these return `ActionController::Parameters` objects rather than plain Ruby hashes, so key-lookup behaviour differs from what the request specs exercise. Both should go through `permit` before being accessed.

app/controllers/course_assigns_controller.rb
```diff
@@ -59,20 +67,23 @@ def set_user_or_team
...
-    when 'two_weeks'
-      DateTime.now + 2.weeks
-    when 'one_month'
-      DateTime.now + 1.month
+  def to_deadline(duration, custom_date = nil)
+    if duration == 'custom'
+      return nil if custom_date.blank?
+
+      begin
+        DateTime.parse(custom_date)
```
dr-r3d: `DateTime.parse` produces a timezone-naive object, inconsistent with `Time.current` used for the relative deadlines in the same method. The application timezone is not applied, so the stored deadline will differ from what the user intended whenever the server timezone differs from UTC.

app/javascript/controllers/course_select_controller.js
```diff
@@ -1,8 +1,46 @@
 import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
+    static targets = ["assignButton", "assignButtonInner", "modal"];
 
     connect() {
         this.showDuration = this.element.dataset.showDuration === "true";
```
dr-r3d: `data-show-duration` was removed from the template, so `this.showDuration` is set to a constant `false` and is never read anywhere in the controller. Dead code.

app/javascript/controllers/course_select_controller.js
```diff
@@ -1,8 +1,46 @@
...
+            if (checkedIds.includes(row.dataset.courseId)) {
+                row.classList.remove("hidden");
+            } else {
+                row.classList.add("hidden");
+            }
+        });
+
+        const container = this.modalTarget.querySelector('[data-deadline-modal-target="courseIdInputs"]');
+        if (container) {
+            container.innerHTML = checkedIds
```
dr-r3d: Building HTML via template-literal interpolation into `innerHTML` is a pattern to avoid even when the values are currently safe integers. If the source of `id` values ever changes, this becomes an XSS vector with no visible change at the call site. Prefer `createElement`/`setAttribute`.

config/locales/en.yml
```diff
@@ -57,6 +57,7 @@ en:
     update: "Update"
     share: "Share"
     choose: "Choose"
+    assign_course: "Assign Course"
```
dr-r3d: `t("button.assign_course")` is called from inside the `neo_component` engine, but the key is defined here in the host app's locale file. The engine should be self-contained — this key belongs in `engines/neo_component/config/locales/en.yml`.

app/javascript/controllers/deadline_modal_controller.js
```diff
@@ -0,0 +1,113 @@
...
+
+    handleSubmitEnd(event) {
+        if (event.detail.success) this.close();
+    }
+
+    handleKeydown(event) {
+        if (event.key !== "Escape") return;
+        const wrapper = this.element.closest(".modal-wrapper");
+        if (wrapper?.style.display !== "none") this.cancel();
```
dr-r3d: Checking `wrapper?.style.display !== "none"` couples the Escape-key handler to the specific inline-style mechanism used to hide the modal. If the hiding strategy changes (e.g. a CSS class toggle), this check will silently stop working. An explicit open/closed flag tracked on the controller would be more robust.

app/javascript/controllers/deadline_modal_controller.js
```diff
@@ -0,0 +1,113 @@
...
+    propagateApplyAllDate() {
+        const dateValue = this.applyAllCustomDateTarget
+            .querySelector('[data-date-picker-target="dateSelector"]')?.value;
+        if (!dateValue) return;
+
+        this.durationSelectTargets.forEach(select => {
+            const row = select.closest('[data-deadline-modal-target="courseRow"]');
+            if (row.classList.contains("hidden") || select.value !== "custom") return;
+
+            const input = row.querySelector('[data-date-picker-target="dateSelector"]');
```
dr-r3d: Querying `[data-date-picker-target="dateSelector"]` from inside this controller reaches into the date-picker controller's private DOM structure. If the date-picker controller ever renames that target, this breaks silently. Consider dispatching a custom event that the date-picker controller handles, or wiring up a Stimulus outlet.

app/javascript/controllers/deadline_modal_controller.js
```diff
@@ -0,0 +1,113 @@
...
+        if (wrapper?.style.display !== "none") this.cancel();
+    }
+
+    toggleApplyAllCustomDate(event) {
+        const duration = event.target.value;
+
+        if (duration === "custom") {
+            this.applyAllCustomDateTarget.classList.remove("hidden");
+            const icon = this.applyAllCustomDateTarget.querySelector('[data-action*="openDatePicker"]');
+            requestAnimationFrame(() => icon?.click());
```
dr-r3d: Auto-clicking the calendar icon via `requestAnimationFrame` is timing-dependent and will break if rendering order changes. Worth checking whether the date-picker component can expose a programmatic open method or a Stimulus outlet action instead.

engines/neo_component/app/views/view_components/modal_component/_modal_box_component.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= turbo_frame_tag 'modal' do %>
 <div data-controller="modal-loader" class="relative">
-    <div class="modal-wrapper" data-controller="modals" data-modals-target="modalBox">     
+    <% wrapper_class = ['modal-wrapper', wrapper_html_options[:class]].compact.join(' ') %>
+    <% extra_wrapper_attrs = wrapper_html_options.except(:class) %>
+    <div class="<%= wrapper_class %>" data-controller="modals" data-modals-target="modalBox" <%= tag.attributes(extra_wrapper_attrs) %>>
```
dr-r3d: Interpolating `<%= tag.attributes(extra_wrapper_attrs) %>` as a raw string directly inside the `div` opening tag is non-standard — the `tag.*` helpers are designed to be used inside `content_tag` or `tag.div` blocks where attribute merging happens automatically. The logic that splits and re-merges `wrapper_html_options` would be cleaner in the Ruby helper, which already builds the locals hash.

engines/neo_component/app/views/view_components/course_select/_list_component.html.erb
```diff
@@ -3,9 +3,17 @@
     <div class="text-notice">No courses to select</div>
   <% else %>
 
-    <ul id="course-list" class="flex flex-col gap-4 overflow-y-auto p-1 h-[700px]">
+    <ul id="course-list" class="flex flex-col gap-4">
```
dr-r3d: The `h-[700px] overflow-y-auto` scroll container was removed. Worth verifying that the page layout still works correctly when a large number of courses is loaded — previously the list was confined to a fixed-height scrollable area.

app/views/course_assigns/_deadline_modal.html.erb
```diff
@@ -0,0 +1,105 @@
+<% modal_footer = capture do %>
```
dr-r3d, line 1: No view spec for this partial. CLAUDE.md requires a view spec for every view — this partial is substantial enough to warrant one covering the Apply-to-All section, per-course rows, and the form structure.

app/controllers/course_assigns_controller.rb
sadrisya: Replaced both raw params.fetch calls with a single params.permit(course_ids: [], durations: {}, custom_dates: {}) at the top of the action, so all three inputs go through strong parameters before use. course_ids was also pulled through the same permitted object for consistency. Specs still green.

app/controllers/course_assigns_controller.rb
sadrisya: swapped DateTime.parse for Time.zone.parse, which applies the app's configured timezone and keeps custom deadlines consistent with the Time.current-based relative deadlines in the same method. Also added TypeError to the rescue clause since Time.zone.parse raises that (rather than ArgumentError) when passed nil. Specs still green.

app/javascript/controllers/course_select_controller.js
sadrisya: Fixed across the board. Removed this.showDuration from course_select_controller.js (and its now-empty connect()). Followed the chain into course_select_item_controller.js — the static outlets, connect(), and toggleDurationSelect (which was the only consumer of showDuration and permanently a no-op) are all gone. Removed the corresponding data-action and outlet attribute from _list_item_component.html.erb too.

app/javascript/controllers/course_select_controller.js
sadrisya: Replaced the innerHTML template-literal approach with createElement + setAttribute + replaceChildren. The values never pass through an HTML parser now, so there's no XSS surface regardless of what id values look like in the future.

app/javascript/controllers/deadline_modal_controller.js
sadrisya: Added this.isOpen = false in connect(), an open() method that sets it to true (and owns the show + focus logic), and close() now sets it back to false. handleKeydown checks this.isOpen — no DOM introspection at all. Wired a Stimulus outlet (deadline-modal) so course_select_controller calls this.deadlineModalOutlet.open() rather than touching style.display directly, keeping all visibility logic in one place.

app/javascript/controllers/deadline_modal_controller.js
sadrisya: Added a handleSetValue listener for a generic date-picker:set-value custom event in date_picker_controller — it updates dateSelectorTarget itself so no outside controller needs to know that target name. In propagateApplyAllDate, replaced the querySelector('[data-date-picker-target=...]') reads: the apply-all value now comes from event.target.value (the input that fired the bubbled change), and each per-course date picker receives a date-picker:set-value dispatch. The deadline_modal_controller only references [data-controller~="date-picker"], which is public.

app/javascript/controllers/deadline_modal_controller.js
sadrisya: Added handleOpen / date-picker:open to date_picker_controller alongside the existing date-picker:set-value pattern. Both toggleApplyAllCustomDate and toggleCustomDate now dispatch date-picker:open to the [data-controller~="date-picker"] element instead of querying [data-action*="openDatePicker"] and simulating a click. The requestAnimationFrame wrapper is kept — showPicker() requires the element to already be visible, so we still need to yield to the browser after removing the hidden class before dispatching the open event.

engines/neo_component/app/views/view_components/modal_component/_modal_box_component.html.erb
sadrisya: Moved all the attribute merging into the Ruby helper: wrapper_class + data-controller/data-modals-target deep-merged with any caller-supplied wrapper_html_options, and modal-content class merged with any caller-supplied html_options. The partial now receives fully-resolved hashes and uses content_tag :div, **wrapper_html_options — no computation or raw tag.attributes interpolation in the template.

engines/neo_component/app/views/view_components/course_select/_list_component.html.erb
sadrisya: Restored overflow-y-auto p-1 h-[700px] on the <ul>. Without it the list grows unbounded and the sticky footer stops behaving correctly once the course count exceeds the viewport.

app/views/course_assigns/_deadline_modal.html.erb
sadrisya: added

config/locales/en.yml
sadrisya: Created engines/neo_component/config/locales/en.yml with both button.cancel and button.assign_course (both were used in the engine's views but only defined in the host app). Added a neo_components.i18n initializer to engine.rb that appends the engine's locale path to config.i18n.load_path, making the engine fully self-contained for its own translation keys.

dr-r3d: This key is now also defined in `engines/neo_component/config/locales/en.yml` (with a different value: `"Review & Assign"` vs `"Assign Course"` here). Both will be loaded, and load order determines which wins — making the displayed label unpredictable. if both are need, use different keys, else remove one

app/views/course_assigns/new.html.erb
```diff
@@ -8,6 +8,10 @@
...
-      show_duration: true,
       submit_path: course_assigns_path
-    ) %>
\ No newline at end of file
+    ) %>
+
+<%= render 'deadline_modal',
+      courses: @courses,
+      search_context: @search_context,
+      submit_path: course_assigns_path %>
```
dr-r3d: File is missing a trailing newline. Most editors and linters expect files to end with `\n`.

dr-r3d (commented):
All 11 previous comments addressed — strong params, `Time.zone.parse`, dead `showDuration`, `createElement` DOM injection, `content_tag` in the modal partial, `isOpen` flag, custom events on the date-picker controller, Stimulus outlet, engine locale file, scroll container preserved, and view spec added. Two remaining issues below.

config/locales/en.yml
sadrisya: Done

app/views/course_assigns/new.html.erb
sadrisya: Done

dr-r3d (approved):

PR #1391, opened 2026-06-05
Task/extend course card for classroom kit

### LINKED ISSUE #1383: Extend course_card_component to support Classroom Kit cards
Extend course_card_component with three optional args so the same component renders both Course and Classroom Kit cards in the Content Studio carousel without duplication.

**Figma:** https://www.figma.com/design/QOquDODFlEr3jKFmyyw96z/AI-Course-Generation?node-id=2699-3954

## Changes

Three backwards-compatible extensions — no existing callers affected:

- `modules_label: 'Lesson'` — new optional arg replacing the hardcoded label in `pluralize(modules_count, 'Lesson')`. Kits pass `'Document'`.
- `duration: nil` — make optional; add `if duration.present?` guard in template so the duration pill is hidden when not provided.
- `enroll_count: nil` — make optional; add `if enroll_count` guard in template so the enroll row is hidden when not provided.

Also adds `studio_kit_card` helper in `content_studio/application_helper.rb` alongside the existing `studio_course_card`.

## Deliverables

- Updated component helper and template
- Usage examples for both the standard course card and the kit card variant
- `ui_manifest.md` updated with the new optional args, their defaults, and a kit card usage example
- PR must include visual diffs showing course cards are unaffected and kit cards render correctly

sadrisya:
Summary
Extends course_card_component with three backwards-compatible optional arguments so it can render both standard Course cards and Classroom Kit cards from a single component without duplication.

Related Issue
Closes [#1383](https://github.com/openvitae-tech/blackboard-lms/issues/1383)

- Changes
- Made duration: optional (nil default) — duration badge conditionally rendered in the template
- Made enroll_count: optional (nil default) — enrol count row + divider hidden when absent
- Added modules_label: 'Lesson' optional param — replaces the hardcoded 'Lesson' string (Kit cards pass 'Document')
- Updated ui_manifest.md with the new optional params and a Classroom Kit usage example
- Added two new sections to the cards gallery — full-featured Course cards and Kit cards side by side for visual comparison
- Fixed cards gallery mobile layout: replaced flex-wrap with a responsive grid (grid-cols-1 sm:grid-cols-2 md:grid-cols-3) to prevent card overlap on small screensSummary
- Extends course_card_component with three backwards-compatible optional arguments so it can render both standard Course cards and Classroom Kit cards from a single component without duplication.

## Screenshots / Visual Diffs

Screenshot 2026-06-08 at 5 19 47 PM
Screenshot 2026-06-08 at 5 20 00 PM

## Checklist


engines/neo_component/spec/helpers/ui_helper_course_card_component_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+
+      it 'renders the divider when enroll_count is present' do
+        doc = render_card(enroll_count: 10)
+        expect(doc.at_css('div.w-\[1px\]')).to be_present
+      end
+
+      it 'omits the enrolment count when nil' do
+        doc = render_card(enroll_count: nil)
+        expect(doc.at_css('svg[data-src*="user"]')).to be_nil
+      end
```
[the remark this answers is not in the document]
sadrisya: replaced it with expect(doc.css('svg').count).to eq(1) — with 
  rating: nil as the default in render_card, only the lessons icon SVG is rendered when enroll_count is       
  absent, making it a genuine assertion.

dr-r3d (commented):
**1. `type_tag` is not documented in `ui_manifest.md`**

The diff adds `modules_label` to the manifest but `type_tag: nil` is absent. Any caller consulting the manifest won't know the param exists.

```diff
# ui_manifest.md — missing line:
+  type_tag: nil,           # nil or Hash with :label, :bg_color, :text_color — renders a coloured footer strip
```

**2. Card width change is a global regression**

`md:w-[312px]` → `md:w-[280px]` is applied to the shared partial, so every existing Course card gets narrower, not just Kit cards. If the Figma spec calls for 280px on Kit cards, a separate width should be passed (or the existing cards should be verified against Figma at the new width before merging).

**3. Animated progress bar affects all cards**

Replacing `bg-primary` with the gradient + animation is applied unconditionally to all cards that have a `progress:` value. This is a visual change to existing course cards in the LMS — scope it explicitly (e.g., `animated_progress: false`) or confirm it's intentional.

**4. `background-size` as a hardcoded inline style**

```erb
style="width: <%= progress %>%; background-size: 200% 100%;"
```

`background-size: 200% 100%` is a static value needed for the animation. Since it can't be expressed as a standard Tailwind utility but is purely presentational, it should live in the animation definition in `tailwind.config.js`, not in the template — avoids mixing concerns and keeps the template clean.

**5. Checklist is fully unchecked**

Neither RSpec nor RuboCop are confirmed to pass. CI status should be green before requesting review; the unchecked boxes signal the PR may not be ready.

sadrisya:
- type_tag missing from manifest — Added type_tag: nil to ui_manifest.md with a note that :text_color is not needed since it's hardcoded as text-grey-300 in the template. Fixed in the latest commit.

- Card width regression — The 280px width is intentional — it's directly from the updated Figma spec for course cards. All course card instances were designed at this width, so it's not a regression but a design-wide update.

- Animated gradient on all cards — Also intentional per Figma. All cards with a progress: value should show the animated gradient — it's a design-wide decision, not scoped to a specific card type.

- background-size inline style — Good catch. Moved background-size: 200% 100% into the gradient-flow keyframe definition in tailwind.config.js so the template no longer carries a mixed-concern inline style. Fixed in the latest commit.

- CI checklist — RSpec (34 examples, 0 failures) and RuboCop (no offenses) both pass. Boxes checked.

engines/content_studio/app/helpers/content_studio/application_helper.rb
```diff
@@ -62,6 +62,18 @@ def studio_course_card(course, _status)
...
+    def studio_kit_card(kit)
+      card = course_card_component(
+        title: kit.title,
+        banner_url: kit.thumbnail_url.presence,
+        modules_count: kit.documents_count,
+        modules_label: 'Document',
+        categories: kit.categories || [],
+        badge: studio_badge(kit.level)
+      )
+      link_to(card, course_structure_path(id: kit.id), class: 'block')
```
dr-r3d: `studio_kit_card` is defined but never called — and it routes to `course_structure_path`, which is the course editor path. Kits will have their own route, so this will silently point at the wrong place once the helper is wired up.

engines/neo_component/app/views/view_components/cards/_course_card_component.html.erb
```diff
@@ -1,46 +1,50 @@
-<div class="group relative items-stretch rounded-xl border border-line-colour bg-white h-full w-auto md:w-[312px]">
+<div class="group relative items-stretch rounded-xl border border-line-colour bg-white h-full w-auto md:w-[280px] text-letter-color">
```
dr-r3d: The card width changed from `md:w-[312px]` to `md:w-[280px]`. This affects every existing caller of `course_card_component`, not just the new Kit variant. Will change this later

engines/neo_component/spec/helpers/ui_helper_course_card_component_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    describe 'duration' do
+      it 'renders the duration badge when present' do
+        doc = render_card(duration: '3h 15m')
+        expect(doc.text).to include('3h 15m')
+      end
+
+      it 'omits the duration badge when nil' do
+        doc = render_card(duration: nil)
+        expect(doc.at_css('div.bg-primary-light-100')).to be_nil
```
dr-r3d: The duration badge no longer uses `bg-primary-light-100` — the template now renders it with `bg-white-light`. This assertion passes vacuously regardless of whether the duration is actually hidden. The test should assert against the rendered text or the class the badge currently uses.

engines/neo_component/spec/helpers/ui_helper_course_card_component_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+      end
+
+      it 'renders the document count label' do
+        doc = render_kit(modules_count: 3)
+        expect(doc.text).to include('3 Documents')
+      end
+
+      it 'hides the duration badge' do
+        doc = render_kit
+        expect(doc.at_css('div.bg-primary-light-100')).to be_nil
```
dr-r3d: Same issue as the duration badge test in the `duration` describe block (line 51): `bg-primary-light-100` is no longer the class used by the duration badge, so this assertion passes vacuously.

engines/neo_component/spec/helpers/ui_helper_course_card_component_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+        stats_row = doc.at_css('div.flex.items-center.justify-between')
+        expect(stats_row).to be_present
+        element_children = stats_row.children.select(&:element?)
+        expect(element_children.last.text).to include('4.5')
+      end
+    end
+
+    describe 'type_tag' do
+      it 'renders the type tag label when present' do
+        doc = render_card(type_tag: { label: 'Course', bg_color: 'bg-primary-light-200', text_color: 'text-white' })
```
dr-r3d, line 149: `text_color` is passed here (and in the other `type_tag` examples below) but the template never reads `type_tag[:text_color]` — it hardcodes `text-letter-colour-medium`. Apply it in the template. set text-letter-colour-medium as default if needed

dr-r3d (commented):
The PR description says a `studio_kit_card(kit)` helper was added - Remove this

engines/neo_component/app/helpers/view_component/card/course_card_component.rb
```diff
@@ -6,15 +6,18 @@ module CourseCardComponent
...
+        modules_label: 'Lesson',
         categories: [],
         rating: nil,
         progress: nil,
         badge: nil,
         description: nil,
-        highlights: []
+        highlights: [],
+        type_tag: nil,
+        card_width: 312
```
dr-r3d: The original issue was an unintentional width change — the correct fix is just reverting the Tailwind class, which is already done above. Adding `card_width` as a new public parameter goes beyond the scope of the fix and introduces functionality no current caller needs. Per the prop-bloat policy, new optional params should only be added when there is an actual call-site that requires them.

engines/neo_component/ui_manifest.md
```diff
@@ -262,15 +262,18 @@ chip_component(
...
+  modules_label: 'Lesson', # override label, e.g. 'Document' for Classroom Kit cards
   categories: [],
   rating: nil,
   progress: nil,
   badge: nil,              # nil or Hash with :label, optional :bg_color, :text_color
   description: nil,
-  highlights: []
+  highlights: [],
+  type_tag: nil            # nil or Hash with :label, :bg_color — renders a full-width coloured footer strip
+                           #   text colour is always text-letter-colour-medium (hardcoded in template)
```
dr-r3d: Two inaccuracies here:\n\n1. The comment still says "text colour is always text-letter-colour-medium (hardcoded in template)" — but the template now uses `type_tag.fetch(:text_color, 'text-letter-colour-medium')`, making `text_color` a supported optional key. The comment should reflect this.\n\n2.

dr-r3d (commented):
Original issues all addressed. Three new issues introduced by the card_width fix — see inline comments.

config/tailwind.config.js
```diff
@@ -46,8 +46,24 @@ module.exports = {
         "slate-grey-light":"#D1D3D4",
         "white":"#FFFFFF",
         "white-light":"#FAFAFA",
+        "primary-light-200":"#AAC9EC",
+        "secondary-light-200":"#B8CB6C",
+        "gold-dark":"#EEC062",
+        "letter-colour-medium":"#4D4D4D",
```
dr-r3d: Naming inconsistency: the existing tokens in this file use American spelling (`letter-color`, `letter-color-light`), but this new key uses British spelling (`letter-colour-medium`). The token family should stay consistent — `letter-color-medium` would match the existing convention.

dr-r3d (approved):

PR #1396, opened 2026-06-09
Update long_course_card_component UI to match Figma design

sadrisya:
## Summary
Update long_course_card_component visuals and mobile layout to match the latest Figma design, add type_tag and publish_status props, and resolve merge conflicts keeping both long_course_card_component and course_card_component doc examples.

Visual changes:

- Title colour: text-primary → text-letter-color (dark grey)
- Stats row: modules/enroll on left, rating shifted right (number-then-star); dividers visible on mobile too
- Duration pill: floating inside image (bottom-2 right-2), all-corner rounded, bg-white-light with border-line-colour-light
- Progress bar: solid bg-primary
- Card radius: rounded-lg mobile / md:rounded-xl desktop
- Image corners: rounded-tl-lg mobile / md:rounded-l-xl desktop
- Content padding: pr-3 py-4 mobile / md:p-5 desktop
- Categories row: mt-auto anchors to card bottom
- TypeTag on mobile: full-width footer bar (md:hidden); desktop: inline pill bottom-right
-

New features:

- type_tag prop: { label:, bg_color: } — renders inline pill (desktop) and footer bar (mobile)
- publish_status prop: 'published' (blue corner diamond) or 'unpublished' (gold corner diamond) with eye icon
- Colour tokens added to tailwind.config.js: grey-dark (#4D4D4D)

**Doc examples added**

- Long Course Card — default, with type tag, with checkbox, with checkbox and type tag, published corner tag, unpublished corner tag

**Spec coverage**

- type_tag rendering and bg_color application
- Rating order by content position (not CSS class)
- publish_status corner tag: published, unpublished, nil, invalid value raises
- md:gap-1 no-op removed; spec selector updated

## Screenshots / Visual Diffs
Screenshot 2026-06-09 at 1 51 20 PM
Screenshot 2026-06-09 at 1 51 31 PM
Screenshot 2026-06-09 at 1 53 17 PM
Screenshot 2026-06-09 at 1 53 25 PM
Screenshot 2026-06-17 at 9 54 49 PM

## Checklist


config/tailwind.config.js
```diff
@@ -46,8 +46,24 @@ module.exports = {
...
+      },
+      keyframes: {
+        'gradient-flow': {
+          '0%':   { backgroundPosition: '0% 0%' },
+          '100%': { backgroundPosition: '-200% 0%' },
+        },
+      },
+      animation: {
+        'gradient-flow': 'gradient-flow 3s linear infinite',
+      },
```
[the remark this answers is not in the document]
sadrisya: the three unrelated Tailwind entries (backgroundImage, keyframes, animation) are removed from this branch. They can be re-added in a separate focused PR when the progress bar feature is ready.

engines/neo_component/app/views/neo_components/ui/cards/index.html.erb
```diff
@@ -48,6 +48,119 @@
...
+          banner_url: nil,
+          duration: '12 Mins',
+          modules_count: '3 Lessons',
+          enroll_count: '60 Enrolled',
+          categories: %w[Safety],
+          checkbox: { name: 'course_ids[]', value: '3' },
+          type_tag: { label: 'Microlesson', bg_color: 'bg-gold-dark', text_color: 'text-white' }
+        ) %>
+  </div>
+<% end %>
```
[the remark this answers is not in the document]
sadrisya: type_tag is already documented in ui_manifest.md at line 291. The finding is invalid — no changes needed.

Skipped — type_tag is already documented in ui_manifest.md line 291 with its expected keys (:label, :bg_color).

config/tailwind.config.js
```diff
@@ -46,6 +46,10 @@ module.exports = {
         "slate-grey-light":"#D1D3D4",
         "white":"#FFFFFF",
         "white-light":"#FAFAFA",
+        "grey-300":"#4D4D4D",
```
dr-r3d: `grey-300` maps to `#4D4D4D`, which is a dark grey (~30% lightness). In Tailwind convention the `300` step is light; the existing tokens here also use semantic names (`slate-grey-light`, `slate-grey-50`) rather than a numeric scale. A name that signals the darkness (e.g. `grey-dark`) would be less likely to mislead future callers.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,9 +1,9 @@
-<div class="group rounded-xl border border-line-colour relative md:max-h-[132px]">
+<div class="group rounded-lg md:rounded-xl border border-line-colour relative">
```
dr-r3d: `md:max-h-[132px]` was on this element in `main` but has been removed here without mention in the PR description. If any caller renders these cards in a fixed-height list view the cards will now grow to content height on desktop. Please confirm the removal is intentional and note it in the description.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -33,80 +33,93 @@
...
-              <%= icon("book", css: "h-4 w-4 text-primary") %>
-              <span class="ml-2 general-text-sm-normal"><%= modules_count %></span>
-            </div>
-            <div class="w-px h-full bg-slate-grey-light"></div>
-            <div class="flex items-center">
-              <%= icon("user", css: "h-4 w-4 text-primary") %>
-              <span class="ml-2 general-text-sm-normal"><%= enroll_count %></span>
+        <div class="flex flex-col gap-3">
+          <div class="flex gap-2 md:gap-4 items-center flex-wrap">
+            <div class="flex gap-1 md:gap-1 items-center">
```
dr-r3d: `md:gap-1` is the same value as `gap-1`, so the responsive prefix has no effect — both breakpoints get the same gap. The class is a no-op and can be dropped. This also matters because the spec (line 190) selects this element by its full class string including `md:gap-1`; anyone cleaning up the redundant class will silently break that test.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -33,80 +33,93 @@
...
-        <div class="hidden md:flex items-center gap-2">
-          <% categories.first(2).each do |category| %>
-            <%= chip_component(text: category, colorscheme: 'primary', icon_name: 'check') %>
-          <% end %>
-          <% overflow = categories.length - 2 %>
-          <% if overflow > 0 %>
-            <span class="general-text-sm-normal text-primary whitespace-nowrap">+<%= overflow %></span>
+          <% if type_tag.present? %>
+            <div class="flex items-center justify-center px-3 py-1 rounded <%= type_tag[:bg_color] %>">
+              <span class="general-text-sm-normal text-grey-300 whitespace-nowrap"><%= type_tag[:label] %></span>
```
dr-r3d: `type_tag[:text_color]` is accepted by callers (and shown in the spec examples) but is never read here — the colour is hardcoded as `text-grey-300` on both this desktop pill and the mobile footer bar (line 122). The manifest correctly documents the hardcoding, but the spec examples still pass `text_color:` as though it is honoured (spec lines 199, 205), which will mislead future callers. Remove `text_color` from the spec hash examples to match what the template actually supports.

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
```diff
@@ -184,6 +184,33 @@ def render_card(**kwargs)
         doc = render_card(rating: nil)
         expect(doc.css('[class*="star"]')).to be_empty
       end
+
+      it 'renders the rating after modules and enroll in the stats row' do
+        doc = render_card(rating: '4.2', modules_count: '6 modules', enroll_count: '50 enrolled')
+        stats_row = doc.at_css('div.flex.gap-1.md\\:gap-1.items-center')
```
dr-r3d: This selector couples the test to the implementation-level class string `flex.gap-1.md:gap-1.items-center`. If the gap value or the redundant `md:gap-1` class is ever changed the test breaks for a non-behavioral reason. Consider asserting the rating order by content position instead of by CSS class.

config/tailwind.config.js
sadrisya: Renamed grey-300 → grey-dark in the tailwind config and updated all 4 usages (2 in the template, 2 in specs)

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: Height is restored

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: Removed text_color from the spec

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: Done — removed md:gap-1 from the template and updated the spec selector to match, then pushed.

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
sadrisya: Done — the test now checks that modules_count, enroll_count, and rating appear in that order in the rendered text, with no coupling to CSS classes.

engines/neo_component/ui_manifest.md
```diff
@@ -324,6 +324,8 @@ long_course_card_component(
   badge: nil,              # nil or Hash with :label, optional :bg_color, :text_color
   description: nil,
   highlights: [],
+  type_tag: nil,           # nil or Hash with :label, :bg_color — inline pill on desktop, footer bar on mobile
+                           #   text colour is always text-grey-300 (hardcoded in template)
```
dr-r3d: Stale token name — the `grey-300` rename to `grey-dark` was done in this branch but this comment was not updated. Should read `text-grey-dark`.

engines/neo_component/ui_manifest.md
```diff
@@ -324,6 +324,8 @@ long_course_card_component(
...
   checkbox: false          # false to hide; or a Hash of check_box_tag options (truthy non-Hash raises ArgumentError)
```
dr-r3d: `publish_status` is not documented here. Please add an entry describing the accepted values (`"published"`, `"unpublished"`, `nil`), the visual effect (corner diamond with eye icon), and that an invalid value raises `ArgumentError`.

engines/neo_component/ui_manifest.md
sadrisya: updated

engines/neo_component/ui_manifest.md
sadrisya: Updated

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -33,80 +33,103 @@
...
-        <div class="hidden md:flex items-center gap-2">
-          <% categories.first(2).each do |category| %>
-            <%= chip_component(text: category, colorscheme: 'primary', icon_name: 'check') %>
-          <% end %>
-          <% overflow = categories.length - 2 %>
-          <% if overflow > 0 %>
-            <span class="general-text-sm-normal text-primary whitespace-nowrap">+<%= overflow %></span>
+          <% if type_tag.present? %>
+            <div class="flex items-center justify-center px-3 py-1 rounded <%= type_tag[:bg_color] %>">
+              <span class="general-text-sm-normal text-grey-dark whitespace-nowrap"><%= type_tag[:label] %></span>
```
dr-r3d: `type_tag[:text_color]` is never read — both the desktop pill (line 108) and the mobile footer bar (line 122) hardcode `text-grey-dark`. Compare with `badge`, which interpolates `badge[:text_color]` directly. A caller who passes `type_tag: { label: "Course", bg_color: "bg-primary-light-200", text_color: "text-white" }` gets no error and silently wrong text colour.

Fix in each spot:
```erb
<span class="general-text-sm-normal <%= type_tag[:text_color].presence || 'text-grey-dark' %> whitespace-nowrap">
```

engines/neo_component/app/helpers/view_component/card/long_course_card_component.rb
```diff
@@ -15,13 +15,20 @@ def long_course_card_component(
         badge: nil,
         description: nil,
         highlights: [],
-        checkbox: false
+        checkbox: false,
+        type_tag: nil,
+        publish_status: nil
```
dr-r3d, line 20: `type_tag` has no shape validation, but `checkbox` (line 22) does. Two failure modes:

1. **Non-Hash truthy value** — `type_tag: 'Course'` reaches the template and calls `'Course'[:bg_color]`, which raises `TypeError: no implicit conversion of Symbol into Integer` — no mention of `type_tag` in the error.
2. **Missing keys** — `type_tag: { label: 'Course' }` (no `:bg_color`) renders `<div class="">` silently.

Follow the `checkbox` pattern:
```ruby
if type_tag && (!type_tag.is_a?(Hash) || type_tag[:label].blank? || type_tag[:bg_color].blank?)
  raise ArgumentError, "type_tag must be nil or a Hash with :label and :bg_color"
end
```

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
```diff
@@ -184,6 +184,32 @@ def render_card(**kwargs)
         doc = render_card(rating: nil)
         expect(doc.css('[class*="star"]')).to be_empty
       end
+
+      it 'renders the rating after modules and enroll in the stats row' do
+        doc = render_card(rating: '4.2', modules_count: '6 modules', enroll_count: '50 enrolled')
+        text = doc.text
+        expect(text.index('6 modules')).to be < text.index('50 enrolled')
+        expect(text.index('50 enrolled')).to be < text.index('4.2')
+      end
```
dr-r3d, line 199: `doc.text` concatenates the entire card's text — title, tooltip, badge, categories, everything. The test passes today only because `'4.2'` doesn't appear anywhere else in the default render. If the title or any other rendered string contained `'4.2'`, `text.index('4.2')` would find it earlier and the assertion would fail or pass for the wrong reason.

Scope to the stats row div instead:
```ruby
stats = doc.at_css('.flex.gap-1.items-center')
text  = stats.text
expect(text.index('6 modules')).to be < text.index('50 enrolled')
expect(text.index('50 enrolled')).to be < text.index('4.2')
```
`.flex.gap-1.items-center` (template line 64) is the div that wraps only the modules/enroll/rating trio.

dr-r3d:
**PR description: new colour tokens overstated**

The description lists four tokens as newly added: `grey-dark`, `primary-light-200`, `secondary-light-200`, `gold-dark`. Only `grey-dark` is actually new in this diff (it is the only line with a `+` prefix in `tailwind.config.js`). The other three were already present in the config before this PR.

Not a code bug, but misleading for anyone using `git log -S` or bisecting to find when a token was introduced — they will land on this PR and find the token was already there.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: The fix applies type_tag[:text_color].presence || 'text-grey-dark' at both locations:

Line 103 — desktop pill (.hidden md:flex)
Line 117 — mobile footer bar (.md:hidden)
This now matches how badge already handles its text_color key, so a caller passing type_tag: { label: 'Course', bg_color: 'bg-primary-light-200', text_color: 'text-white' } will get the correct colour instead of silently falling back to text-grey-dark.

engines/neo_component/app/helpers/view_component/card/long_course_card_component.rb
sadrisya: Added shape validation for type_tag in long_course_card_component

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
sadrisya: Partial — added checkbox_opts[:id] = nil unless checkbox_opts.key?(:id) so Rails doesn't auto-generate an id when none is supplied, preserving the original behaviour and keeping ids unique across multiple cards
Spec — replaced the outdated class expectations (cursor-pointer, extra-class) with a presence check and 'hidden' — reflecting that input_checkbox_component now owns the input's class

sadrisya:
> **PR description: new colour tokens overstated**
> 
> The description lists four tokens as newly added: `grey-dark`, `primary-light-200`, `secondary-light-200`, `gold-dark`. Only `grey-dark` is actually new in this diff (it is the only line with a `+` prefix in `tailwind.config.js`). The other three were already present in the config before this PR.
> 
> Not a code bug, but misleading for anyone using `git log -S` or bisecting to find when a token was introduced — they will land on this PR and find the token was already there.

Done

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -33,80 +29,103 @@
...
-        <div class="hidden md:flex items-center gap-2">
-          <% categories.first(2).each do |category| %>
-            <%= chip_component(text: category, colorscheme: 'primary', icon_name: 'check') %>
-          <% end %>
-          <% overflow = categories.length - 2 %>
-          <% if overflow > 0 %>
-            <span class="general-text-sm-normal text-primary whitespace-nowrap">+<%= overflow %></span>
+          <% if type_tag.present? %>
+            <div class="flex items-center justify-center px-3 py-1 rounded <%= type_tag[:bg_color] %>">
+              <span class="general-text-sm-normal <%= type_tag[:text_color].presence || 'text-grey-dark' %> whitespace-nowrap"><%= type_tag[:label] %></span>
```
dr-r3d, line 105: The manifest documents `type_tag` as accepting only `:label` and `:bg_color`, and states "text colour is always text-grey-dark (hardcoded in template)." But the template supports an optional `:text_color` key that silently overrides the default. This undocumented API surface also goes unvalidated by `validate_type_tag_arg!` — a caller can pass any arbitrary class string via `:text_color` with no error. Either remove the lookup and hardcode the class to match the manifest, or update the manifest and validator to acknowledge `:text_color` as an optional key.

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
```diff
@@ -71,11 +71,11 @@ def render_card(**kwargs)
...
       end
 
-      it 'merges a caller-supplied class with the base class instead of raising' do
+      it 'does not raise when a caller supplies a class and renders the checkbox input' do
         doc = render_card(checkbox: { name: 'course_ids[]', value: '1', class: 'extra-class' })
         input = doc.at_css('input[type="checkbox"]')
-        expect(input['class']).to include('extra-class')
-        expect(input['class']).to include('cursor-pointer')
+        expect(input).to be_present
+        expect(input['class']).to include('hidden')
```
dr-r3d, line 82: The previous assertion verified that a caller-supplied `class` was forwarded to the checkbox input. That contract is now silently broken — `InputCheckboxComponent#initialize` unconditionally overwrites `html_options[:class]` with `"hidden peer"`, so extra classes never reach the input element. The replacement assertion (`include("hidden")`) passes trivially and no longer tests the forwarding behaviour. The test description ("does not raise when a caller supplies a class") is accurate, but the original coverage is gone without any indication of why. Worth a comment in the spec documenting what actually happens to a caller-supplied class.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,17 +1,13 @@
-<div class="group rounded-xl border border-line-colour relative md:max-h-[132px]">
+<div class="group rounded-lg md:rounded-xl border border-line-colour relative md:max-h-[132px] w-full">
   <% if checkbox %>
     <% checkbox_opts = checkbox.is_a?(Hash) ? checkbox.dup : {} %>
-    <% checkbox_id    = checkbox_opts.delete(:id) %>
-    <% checkbox_class = ['w-5 h-5 rounded border border-slate-grey-light checked:bg-primary checked:border-primary focus:ring-0 cursor-pointer', checkbox_opts.delete(:class)].compact.join(' ') %>
-    <span class="absolute top-4 right-4 z-10"
+    <% cb_name  = checkbox_opts.delete(:name) || 'checkbox' %>
+    <% cb_value = checkbox_opts.delete(:value) %>
+    <% checkbox_opts[:id] = nil unless checkbox_opts.key?(:id) %>
```
dr-r3d, line 7: `id: nil` renders `id=""` in the HTML — an empty string attribute, not the absence of one. If the intent here is to suppress the auto-generated ID that `check_box_tag` derives from the field name, `nil` may not achieve that reliably across all Rails tag helpers.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: - engines/neo_component/ui_manifest.md — updated type_tag docs for long_course_card_component to document :text_color as optional and clarify the default, matching the course_card_component entry on line 308.
- engines/neo_component/app/helpers/view_component/card/long_course_card_component.rb — validator now rejects any keys outside %i[label bg_color text_color], so arbitrary strings can no longer be passed silently.
- spec file — added three new cases: :text_color applied when supplied, defaults to text-grey-dark when omitted, unknown keys raise ArgumentError.

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
sadrisya: - Renamed the example to "renders the checkbox input even when a caller supplies a class option" — removes the misleading implication that forwarding is tested.
- Added a comment explaining that InputCheckboxComponent overwrites :class with "hidden peer", so caller-supplied classes are silently dropped by design.
- Added expect(input['class']).not_to include('extra-class') to make the actual behaviour an explicit assertion rather than a silent gap.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: check_box_tag with id: nil omits the attribute entirely, it does not render id="". Confirmed with bin/rails runner. Added an inline comment to the template to make the intent explicit.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -33,80 +30,103 @@
...
-      <div class="absolute top-0 left-0 z-10 px-2 py-1 general-text-sm-medium <%= badge[:bg_color] %> <%= badge[:text_color] %> rounded-tl-xl rounded-br-[8px] border-b border-r border-line-colour">
+      <div class="absolute top-0 left-0 z-10 px-2 py-1 general-text-sm-medium <%= badge[:bg_color] %> <%= badge[:text_color] %> rounded-tl-lg md:rounded-tl-xl rounded-br-[8px] border-b border-r border-line-colour">
         <%= badge[:label] %>
       </div>
     <% end %>
 
     <div class="relative self-stretch">
       <% if banner_url.present? %>
-        <%= image_tag banner_url, class: 'w-[100px] md:w-[297px] h-full md:h-[130px] rounded-l-xl object-cover' %>
+        <%= image_tag banner_url, class: 'w-[100px] md:w-[297px] h-full md:h-[130px] rounded-tl-lg md:rounded-l-xl object-cover' %>
```
dr-r3d: The image and placeholder both use `rounded-tl-lg md:rounded-l-xl`, which only rounds the top-left corner on mobile. The old code used `rounded-l-xl` (both left corners on every breakpoint). The outer card has `rounded-lg` but no `overflow-hidden`, so CSS does not clip children to the card's border-radius — the image's square bottom-left corner will be visible against the card's rounded border on mobile. The placeholder div at line 42 has the same problem.

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
```diff
@@ -184,6 +189,53 @@ def render_card(**kwargs)
...
+
+      it 'applies the supplied bg_color class to the type tag' do
+        tag = { label: 'Classroom Kit', bg_color: 'bg-secondary-light-200' }
+        doc = render_card(type_tag: tag)
+        expect(doc.at_css('div.bg-secondary-light-200')).to be_present
+      end
+
+      it 'omits the type tag when nil' do
+        doc = render_card(type_tag: nil)
+        expect(doc.css('div.bg-primary-light-200, div.bg-secondary-light-200, div.bg-gold-dark')).to be_empty
```
dr-r3d: The negative assertion uses a closed set of three bg-color class names (`bg-primary-light-200`, `bg-secondary-light-200`, `bg-gold-dark`) to verify the type_tag is absent. A type_tag rendered with any other bg_color (e.g. `bg-gold`, `bg-danger`, or a new token added in a future doc example) would silently pass this test even though the tag is present. A structural marker on the type_tag element would give a reliable negative assertion regardless of bg_color.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: fixed

engines/neo_component/spec/helpers/ui_helper_long_course_card_component_spec.rb
sadrisya: added data-testid="type-tag" to both type tag elements (desktop inline pill and mobile bottom bar) in the template. Updated the negative assertion in the spec to use doc.at_css('[data-testid="type-tag"]') instead of the closed set of bg-color class names, so the assertion holds for any bg_color value, current or future.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -1,17 +1,14 @@
-<div class="group rounded-xl border border-line-colour relative md:max-h-[132px]">
+<div class="group rounded-lg md:rounded-xl border border-line-colour relative md:max-h-[132px] w-full overflow-hidden">
```
dr-r3d: Adding `overflow-hidden` to this container clips the description hover tooltip. The tooltip (`absolute left-0 bottom-full z-50 mb-3 w-80`) is a child of this element and is positioned above the card using `bottom-full`. Because this div has both `position: relative` (making it the tooltip's containing block) and `overflow: hidden`, the browser clips everything that extends outside the card's box — the tooltip included. Any card rendered with a `description:` will show no popup on hover.

also add description to cards examples, so that it is easy to verify this

engines/neo_component/ui_manifest.md
```diff
@@ -324,6 +324,13 @@ long_course_card_component(
...
+  type_tag: nil,           # nil or Hash with :label, :bg_color, optional :text_color — inline pill on desktop, footer bar on mobile
+                           #   :text_color defaults to text-grey-dark when omitted
+  publish_status: nil,     # nil | 'published' | 'unpublished'
+                           #   nil — no indicator rendered
+                           #   'published'   — corner diamond (top-right) with eye icon, blue background (bg-primary-light-200)
+                           #   'unpublished' — corner diamond with eye icon, gold background (bg-gold)
+                           #   Any other value raises ArgumentError
   checkbox: false          # false to hide; or a Hash of check_box_tag options (truthy non-Hash raises ArgumentError)
                            #   Hash keys: :name (field name), :value, :checked (Boolean), :id,
                            #   plus any other HTML attributes (e.g. data-*, class overrides).
```
dr-r3d: The manifest still documents "class overrides" as a supported key in the `checkbox:` hash, but this is no longer true after switching to `input_checkbox_component`. `InputCheckboxComponent#initialize` unconditionally does `self.html_options[:class] = 'hidden peer'` (line 19), overwriting any `:class` the caller passes without merging. The old `check_box_tag` path explicitly merged caller classes; this one does not. The manifest should be corrected so callers do not pass `:class` expecting it to take effect.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
```diff
@@ -33,80 +30,103 @@
...
-
-        <div class="hidden md:flex items-center gap-2">
-          <% categories.first(2).each do |category| %>
-            <%= chip_component(text: category, colorscheme: 'primary', icon_name: 'check') %>
-          <% end %>
-          <% overflow = categories.length - 2 %>
-          <% if overflow > 0 %>
-            <span class="general-text-sm-normal text-primary whitespace-nowrap">+<%= overflow %></span>
+          <% if type_tag.present? %>
+            <div class="flex items-center justify-center px-3 py-1 rounded <%= type_tag[:bg_color] %>" data-testid="type-tag">
```
dr-r3d, line 104: `data-testid="type-tag"` appears on both the desktop pill (line 104) and the mobile footer bar (line 118), but no spec in this PR queries `[data-testid="type-tag"]` — all assertions target CSS classes directly (e.g. `div.bg-secondary-light-200`). If this attribute was added for testing purposes, it can be removed from both elements.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: overflow-hidden on the outer container clips the tooltip because it's a child positioned with bottom-full. Each inner element already carries its own border-radius, so removing overflow-hidden from the outer div is safe.

engines/neo_component/ui_manifest.md
sadrisya: The manifest now explicitly warns callers that :class is silently overwritten by input_checkbox_component and removes the "class overrides" claim.

engines/neo_component/app/views/view_components/cards/_long_course_card_component.html.erb
sadrisya: The data-testid="type-tag" on both elements is intentional and used by spec line 217 (ui_helper_long_course_card_component_spec.rb:217) to assert absence when type_tag: nil. Removing it would break that assertion. The PR comment's premise is incorrect — the attribute is queried.

dr-r3d (approved):

PR #1436, opened 2026-06-24
Fix kit progress banner visibility and reuse shared progress_card partial

### LINKED ISSUE #1424: [Bug] title "Progress banner disappears instantly on generation complete — 100% state never visible
## What
Progress banner disappears instantly when kit generation completes — the 100% state is never visible to the user.

## Steps to Reproduce
1.Start a new kit generation with a single component selected.
2.Watch the progress banner during generation.
3.Generation completes — banner vanishes with no visible 100% state.

## Expected Behavior
When generation reaches 100%, the banner should stay visible for ~1.5 seconds so the user can see the completed state before it hides.

## Actual Behavior
The banner hides immediately the moment generation completes. With a single component it jumps from 0% straight to hidden — the user may not even realise it was a progress bar. With multiple components at least intermediate states (50%, 75%) are visible, but the final 100% is still never shown.

## Area Affected
Content Studio — Classroom Kit generation flow

## Agent Instructions

- File: engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
- The pendingValueChanged callback hides the banner as soon as pending becomes false with no delay.
- The previousPending === undefined guard must be kept — it prevents the banner flashing when a user directly opens an already-generated kit.
- The fix should only hold the banner visible when the transition happens during the current page lifecycle (i.e. previousPending === true), then hide after ~1500 ms.
- Related view: engines/content_studio/app/views/content_studio/classroom_kits/structure/show.html.erb

sadrisya:
- Replace inline banner in classroom_kits/structure/show with the shared content_studio/shared/_progress_card partial (already used by courses)
- Extend _progress_card to accept optional banner_target local so the Stimulus target attribute can be applied from the caller
- Track _startedPending on connect() so the banner is hidden immediately when opening an already-complete kit, shown throughout generation when opened from in-progress, and stays visible at 100% until the user refreshes the page

**Summary**
Fix classroom kit progress banner visibility logic and replace the inline banner with the shared _progress_card partial already used by courses.

**Related Issue**
Closes [#1424](https://github.com/openvitae-tech/blackboard-lms/issues/1424)

**Changes**

- Reuse shared partial — replaced the 20-line inline banner in classroom_kits/structure/show.html.erb with render 'content_studio/shared/progress_card', eliminating duplication with the courses flow
- Extended _progress_card — added optional banner_target local (defaults to false) so the kit can wire up the Stimulus data-structure-polling-target="banner" attribute without affecting the existing courses usage
- Fixed banner visibility logic in structure_polling_controller.js — introduced this._startedPending on connect() to track the page's initial pending state:
- Opened from Completed → banner hidden immediately via display: none, never shown
- Opened from In Progress → banner visible throughout generation; stays visible at 100% after completion
- Refresh after completion → connect() re-fires, _startedPending = false, banner hidden again
- Removed BANNER_HIDE_DELAY_MS — auto-hide timer eliminated; banner now persists at 100% until the user navigates away or refreshesSummary
- Fix classroom kit progress banner visibility logic and replace the inline banner with the shared _progress_card partial already used by courses.

## Checklist

### DIFF
--- a/engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
+++ b/engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
@@ -29,7 +29,7 @@ export default class extends Controller {
           this.bannerTarget.style.display = ''
           if (prevSource === 'generation') {
             // Hide after 1 second for generation flow
-            setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
+            this.hideTimer = setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
           }
           // in_progress: stays visible until next page refresh
         } else {
@@ -63,6 +63,7 @@ export default class extends Controller {
 
   disconnect() {
     clearTimeout(this.timer)
+    clearTimeout(this.hideTimer)
     window.removeEventListener('module-select:drag-start', this._onDragStart)
     window.removeEventListener('module-select:drag-end', this._onDragEnd)
     document.removeEventListener('turbo:before-visit', this._onBeforeVisit)
@@ -76,7 +77,7 @@ export default class extends Controller {
     sessionStorage.removeItem(this._storageKey)
     this.bannerTarget.style.display = ''
     if (this.sourceValue === 'generation') {
-      setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
+      this.hideTimer = setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
     }
   }
 
--- a/engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
+++ b/engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
@@ -25,7 +25,12 @@ export default class extends Controller {
         const prevSource = sessionStorage.getItem(this._storageKey)
         if (prevSource) {
           // Frame reloaded after reaching 100%
-          this._showBanner(prevSource)
+          sessionStorage.removeItem(this._storageKey)
+          this.bannerTarget.style.display = ''
+          if (prevSource === 'generation') {
+            // Hide after 1 second for generation flow
+            this.hideTimer = setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
+          }
           // in_progress: stays visible until next page refresh
         } else {
           this.bannerTarget.style.display = 'none'
@@ -69,7 +74,11 @@ export default class extends Controller {
   pendingValueChanged(pending, previousPending) {
     if (pending || previousPending === undefined || !this.hasBannerTarget) return
     if (!this._startedPending || this.sourceValue === 'completed') return
-    this._showBanner(this.sourceValue)
+    sessionStorage.removeItem(this._storageKey)
+    this.bannerTarget.style.display = ''
+    if (this.sourceValue === 'generation') {
+      this.hideTimer = setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
+    }
   }
 
   thumbnailUrlValueChanged(url) {
@@ -84,14 +93,6 @@ export default class extends Controller {
     img.src = url
     img.classList.replace('object-contain', 'object-cover')
     img.dataset.loadedUrl = url
-  }
-
-  _showBanner(source) {
-    sessionStorage.removeItem(this._storageKey)
-    this.bannerTarget.style.display = ''
-    if (source === 'generation') {
-      this.hideTimer = setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
-    }
   }
 
   _schedulePoll() {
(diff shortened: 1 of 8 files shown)


engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
```diff
@@ -2,17 +2,41 @@ import { Controller } from "@hotwired/stimulus"
...
+          this.bannerTarget.style.display = ''
+          if (prevSource === 'generation') {
+            // Hide after 1 second for generation flow
+            setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
+          }
+          // in_progress: stays visible until next page refresh
+        } else {
+          this.bannerTarget.style.display = 'none'
+        }
+      }
```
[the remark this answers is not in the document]
sadrisya: cleared _storageKey from sessionStorage inside pendingValueChanged after the banner logic runs.

engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
```diff
@@ -2,17 +2,41 @@ import { Controller } from "@hotwired/stimulus"
...
+    }
```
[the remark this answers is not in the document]
sadrisya: both setTimeout calls now store their reference in this.hideTimer, and disconnect() clears it alongside this.timer.

engines/content_studio/app/views/content_studio/shared/_progress_card.html.erb
```diff
@@ -1,6 +1,8 @@
-<% completed = local_assigns.fetch(:completed, 0) %>
-<% total = local_assigns.fetch(:total, 0) %>
-<div class="max-w-2xl mx-auto w-full bg-white border border-secondary-dark rounded-lg overflow-hidden">
+<% completed     = local_assigns.fetch(:completed, 0) %>
+<% total         = local_assigns.fetch(:total, 0) %>
+<% banner_target = local_assigns.fetch(:banner_target, false) %>
+<div class="max-w-2xl mx-auto w-full bg-white border border-secondary-dark rounded-lg overflow-hidden"
+     <% if banner_target %>data-structure-polling-target="banner"<% end %>>
```
dr-r3d: The partial now embeds `data-structure-polling-target="banner"` directly, which means it knows the name of a specific Stimulus controller. If a second caller ever needs to target this partial from a different controller, the only options are to add more conditionals or fork the partial.

A cleaner approach is to accept an `html_options` hash and merge it onto the root div, so the caller wires up whatever target it needs without the partial knowing anything about controllers:

```erb
<%# _progress_card.html.erb %>
<% html_options = local_assigns.fetch(:html_options, {}) %>
<div class="max-w-2xl mx-auto w-full bg-white border border-secondary-dark rounded-lg overflow-hidden"
     <%= tag.attributes(html_options) %>>
```

```erb
<%# show.html.erb %>
<%= render 'content_studio/shared/progress_card',
      ...,
      html_options: { data: { 'structure-polling-target': 'banner' } } %>
```

This keeps the partial controller-agnostic and lets any caller attach whatever data attributes it needs.

engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
```diff
@@ -2,17 +2,41 @@ import { Controller } from "@hotwired/stimulus"
...
+      if (source === 'completed') {
+        this.bannerTarget.style.display = 'none'
+      } else if (this._startedPending) {
+        sessionStorage.setItem(this._storageKey, source || 'generation')
+      } else {
+        const prevSource = sessionStorage.getItem(this._storageKey)
+        if (prevSource) {
+          // Frame reloaded after reaching 100%
+          sessionStorage.removeItem(this._storageKey)
+          this.bannerTarget.style.display = ''
```
dr-r3d: The same three-step sequence — `sessionStorage.removeItem` → `style.display = ''` → conditional `setTimeout` hide — appears identically in `connect()` (lines 28–33) and `pendingValueChanged()` (lines 77–81), differing only in the variable holding the source string (`prevSource` vs `this.sourceValue`).

If the delay, hide condition, or display property ever changes, both blocks need updating. They can be unified into a private method:

```js
_showBanner(source) {
  sessionStorage.removeItem(this._storageKey)
  this.bannerTarget.style.display = ''
  if (source === 'generation') {
    this.hideTimer = setTimeout(() => { this.bannerTarget.style.display = 'none' }, 1000)
  }
}
```

Then the two call sites each become a single line.

engines/content_studio/app/views/content_studio/shared/_progress_card.html.erb
sadrisya: Done. The partial no longer knows about any Stimulus controller — it just merges whatever html_options the caller provides onto the root div. The courses caller is unaffected since it never passed banner_target.

engines/content_studio/app/javascript/content_studio/controllers/structure_polling_controller.js
sadrisya: Done. The two duplicated blocks are now a single _showBanner(source) call each, and the shared logic lives in one place.

dr-r3d (approved):

PR #1440, opened 2026-06-25
Restore active tab on back-navigation from course/kit structure

### LINKED ISSUE #1439: Recent Creations tab resets to "In Progress" after navigating back from course/kit structure
**Summary**
In Content Studio, the Recent Creations section has two tabs — In Progress and Completed. When a user clicks on a course or classroom kit from the Completed tab, views its structure page, and then presses the back button, they are returned to the index page with the In Progress tab active instead of Completed. The same applies in reverse — any tab selection is lost on back-navigation.

**Steps to Reproduce**

1. Go to Content Studio (/content-studio)
2. In the Recent Creations section, click the Completed tab
3. Click on any course or classroom kit card
4. On the structure page, press the back (←) button
5. Observe — the In Progress tab is selected instead of Completed

**Expected Behaviour**
Pressing the back button from a structure page should return the user to the Content Studio index with the same tab they navigated from still active.

sadrisya:
When a user opens a course or classroom kit from the Content Studio Recent Creations section and presses the back button, the index page now re-activates the tab (In Progress or Completed) they came from instead of always defaulting to In Progress.

- Pass `tab` param in course/kit structure URLs (in_progress/completed)
- Back buttons on both structure pages preserve the tab param via root_path(tab:)
- auto_click_controller gains an `index` value so it can activate any tab on connect
- Content Studio index reads params[:tab] and passes the correct index to auto-click

## Summary
Fixes a bug where navigating back from a course or classroom kit structure page always reset the Recent Creations section to the "In Progress" tab, regardless of which tab the user came from.

## Related Issue
Closes #1439

## Changes
- `auto_click_controller.js` — added `index` value (default `0`); on connect now clicks `tabItemTargets[indexValue]` instead of always the first tab
- `application_helper.rb` — `studio_course_card` and `studio_kit_card` now embed a `tab` param in their structure URL links (`in_progress` or `completed`)
- `courses/structure/show.html.erb` — back button updated to `root_path(tab: params[:tab])` to carry the tab context back
- `classroom_kits/structure/show.html.erb` — same back button fix as course structure
- `courses/index.html.erb` — Recent Creations section reads `params[:tab]`, computes active tab index, and passes it to `auto-click` controller via `data-auto-click-index-value`

## Screenshots / Visual Diffs

https://github.com/user-attachments/assets/85e7c381-14c8-4a94-9f8c-9744856ec283

## Checklist

### DIFF
--- a/engines/content_studio/app/helpers/content_studio/application_helper.rb
+++ b/engines/content_studio/app/helpers/content_studio/application_helper.rb
@@ -73,7 +73,7 @@ module ContentStudio
       end
     end
 
-    def studio_course_card(course, _status)
+    def studio_course_card(course, status = nil)
       card = course_card_component(
         title: course.title,
         banner_url: course.thumbnail_url.presence,
@@ -86,10 +86,10 @@ module ContentStudio
         badge: studio_badge(course.level),
         type_tag: { label: 'Course', bg_color: 'bg-primary-light-200', text_color: 'text-letter-color' }
       )
-      link_to(card, course_structure_path(id: course.id), class: 'block')
+      link_to(card, course_structure_path(id: course.id, tab: status), class: 'block')
     end
 
-    def studio_kit_card(kit)
+    def studio_kit_card(kit, source = nil)
       card = course_card_component(
         title: kit.title.presence || 'Untitled Kit',
         banner_url: kit.thumbnail_url.presence,
@@ -98,7 +98,8 @@ module ContentStudio
         categories: [],
         type_tag: { label: 'Classroom Kit', bg_color: 'bg-secondary-light-200', text_color: 'text-letter-color' }
       )
-      link_to(card, kit_structure_path(id: kit.id), class: 'block')
+
+      link_to(card, kit_structure_path(id: kit.id, source: source, tab: source), class: 'block')
     end
 
     def lesson_status_label(status)
--- a/engines/content_studio/app/helpers/content_studio/application_helper.rb
+++ b/engines/content_studio/app/helpers/content_studio/application_helper.rb
@@ -73,7 +73,7 @@ module ContentStudio
       end
     end
 
-    def studio_course_card(course, status = nil)
+    def studio_course_card(course, tab = nil)
       card = course_card_component(
         title: course.title,
         banner_url: course.thumbnail_url.presence,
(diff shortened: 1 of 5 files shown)


engines/content_studio/app/helpers/content_studio/application_helper.rb
```diff
@@ -98,7 +98,8 @@ def studio_kit_card(kit)
         categories: [],
         type_tag: { label: 'Classroom Kit', bg_color: 'bg-secondary-light-200', text_color: 'text-letter-color' }
       )
-      link_to(card, kit_structure_path(id: kit.id), class: 'block')
+
+      link_to(card, kit_structure_path(id: kit.id, source: source, tab: source), class: 'block')
```
dr-r3d: `source: source` and `tab: source` are set to the same value, but only `params[:tab]` is ever read — by the back-link on the kit structure page and by the index to compute `active_tab_index`. `params[:source]` has no consumers anywhere in the engine, so it is dead weight in every kit card URL.

engines/content_studio/app/views/content_studio/courses/index.html.erb
```diff
@@ -71,19 +71,23 @@
     <div class="flex flex-col gap-3">
       <p class="heading-xl text-primary px-2">Recent Creations</p>
 
+      <% active_tab_index = params[:tab] == 'completed' ? 1 : 0 %>
```
dr-r3d: `params[:tab] == 'completed' ? 1 : 0` hardcodes the tab position as a numeric index. If a third tab is inserted before "Completed", this ternary still returns `1` and the wrong tab is restored — silently, with no error. The numeric index needs to stay in sync with the DOM order of `data-auto-click-target="tabItem"` elements but nothing enforces that relationship.

engines/content_studio/app/helpers/content_studio/application_helper.rb
sadrisya: Removed the redundant source: source — the URL now passes only tab: source, which is what the kit structure page actually reads.

engines/content_studio/app/views/content_studio/courses/index.html.erb
sadrisya: tab_order is the single source of truth — inserting a new tab means adding it to both the array and the DOM in the same position, and Array#index derives the numeric value automatically. An unrecognised params[:tab] still falls back to 0.

engines/content_studio/app/helpers/content_studio/application_helper.rb
```diff
@@ -98,7 +98,8 @@ def studio_kit_card(kit)
         categories: [],
         type_tag: { label: 'Classroom Kit', bg_color: 'bg-secondary-light-200', text_color: 'text-letter-color' }
       )
-      link_to(card, kit_structure_path(id: kit.id), class: 'block')
+
```
dr-r3d: `studio_course_card` names this param `status`; `studio_kit_card` names it `source`. Both represent the same thing — which tab the card came from — but the inconsistent naming makes it harder to grep for usages and understand the intent at call sites.

sadrisya: Both methods now use tab consistently. The source: tab in studio_kit_card is preserved since the structure view's polling controller reads params[:source] separately.

dr-r3d (approved):

PR #1441, opened 2026-06-25
Fix back button always showing on admin My Courses page

sadrisya:
Back button on the admin courses list should only appear when navigating from My Content, not when arriving directly via the sidebar.

## Summary
Conditionally render the back button on the admin My Courses page based on navigation source.
## Related Issue
Closes #

## Changes

- Added from: 'content' query param to the Courses link in app/views/content/index.html.erb
- Wrapped the back button in app/views/courses/list/_admin.html.erb with a condition so it only renders when params[:from] == 'content'

## Checklist

### DIFF
--- a/app/views/content/index.html.erb
+++ b/app/views/content/index.html.erb
@@ -4,7 +4,7 @@
 
 <div class="flex flex-col">
   <% if policy(:course).manage? %>
-    <%= link_to manage_courses_path, class: "flex items-center gap-3 border-b border-line-colour pl-2 pr-4 py-5 hover:bg-primary-light-50" do %>
+    <%= link_to manage_courses_path(from: 'content'), class: "flex items-center gap-3 border-b border-line-colour pl-2 pr-4 py-5 hover:bg-primary-light-50" do %>
       <span class="flex-1 text-base font-medium text-primary"><%= t('content.courses') %></span>
       <div class="flex items-center gap-2">
         <%= chip_component(text: "#{@courses_published_count} #{t('content.published')}", icon_name: 'eye', colorscheme: 'published', pill: true) %>
--- a/app/views/content/index.html.erb
+++ b/app/views/content/index.html.erb
@@ -4,7 +4,7 @@
 
 <div class="flex flex-col">
   <% if policy(:course).manage? %>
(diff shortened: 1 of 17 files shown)


app/views/courses/list/_admin.html.erb
```diff
@@ -3,7 +3,7 @@
     <%= render 'filters', tags: tags, filter_url: manage_courses_path %>
     <div class="flex flex-row justify-between">
       <div class="flex items-center gap-3">
-        <%= render 'shared/components/back_button', back_link: content_path %>
+        <%= render 'shared/components/back_button', back_link: content_path if params[:from] == 'content' %>
```
dr-r3d: The `from=content` param is not forwarded through the filter form or search form — both are plain GET forms with no hidden field for `from`, so applying a filter or searching generates a URL without `?from=content` and the back button disappears mid-session for privileged users who arrived from the content page.

Since admins never have access to the content page, a simpler condition would be `unless current_user.is_admin?` — no URL param needed, no forwarding concern, and the button is consistently present for privileged users regardless of how they navigate within the page.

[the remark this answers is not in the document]
sadrisya: app/views/courses/list/_admin.html.erb — added hidden_field_tag :from inside the search form so searching preserves params[:from]
app/views/courses/_filters.html.erb — added the same hidden field inside the filter form

sadrisya: The back button now shows unconditionally for non-admin users (privileged users who have access to the content page) and the hidden-field forwarding is gone entirely — net result is simpler and more robust.

app/views/content/index.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="flex flex-col">
   <% if policy(:course).manage? %>
-    <%= link_to manage_courses_path, class: "flex items-center gap-3 border-b border-line-colour pl-2 pr-4 py-5 hover:bg-primary-light-50" do %>
+    <%= link_to manage_courses_path(from: 'content'), class: "flex items-center gap-3 border-b border-line-colour pl-2 pr-4 py-5 hover:bg-primary-light-50" do %>
```
dr-r3d: `from: 'content'` is no longer read anywhere — the back button condition was changed to `unless current_user.is_admin?`, so `params[:from]` has no consumers. This param can be removed from the link.

dr-r3d (approved):

PR #1452, opened 2026-06-30
Add kind discriminator column to courses table for microlesson support

### LINKED ISSUE #1442: Migration: add kind column to courses table
Add a `kind` discriminator to `courses` to distinguish microlessons from regular courses.

## Changes
- Migration: `add_column :courses, :kind, :string, null: false, default: 'course'` + index
- `Course` model: `enum :kind, { course: 'course', microlesson: 'microlesson' }`
- Add a default scope or named scope to exclude microlessons from existing course listings

## Notes
- `default: 'course'` means no backfill needed — all existing rows get `'course'` automatically
- Unblocks: Internal API + MicrolessonSaveService

sadrisya:
- Migration: add kind string column (null: false, default: 'course') with index
- Course model: add kind enum (course | microlesson) with :kind prefix
- Add courses_only scope to exclude microlessons from standard course queries

## Summary
- Adds `kind` string column to `courses` table (`null: false, default: 'course'`) with an index
- Adds `enum :kind` to the `Course` model with two values: `course` and `microlesson` (prefixed to avoid method name conflicts)
- Adds `courses_only` named scope to explicitly exclude microlessons from standard course queries

## Related Issue
Closes #1442

## Changes

- db/migrate/20260630094148_add_kind_to_courses.rb — migration adding kind column with default 'course' and index
- db/schema.rb — auto-updated to reflect the new column
- app/models/course.rb — added enum :kind and scope :courses_only-
- 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for distinguishing between standard courses and microlessons.
  * Course records now default to the standard course type, and the app can filter to show only full courses.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -8,6 +8,9 @@ class Course < ApplicationRecord
   PER_PAGE_LIMIT = 12
 
   enum :visibility, { public: 'public', private: 'private' }, prefix: :visibility
+  enum :kind, { course: 'course', microlesson: 'microlesson' }, prefix: :kind
+
+  scope :courses_only, -> { where(kind: :course) }
 
   belongs_to :learning_partner, optional: true
 
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -27,7 +27,7 @@ class CoursesController < ApplicationController
                                        .includes(:tags, banner_attachment: :blob)
       @courses = @courses.page(filter_params[:page])
     else
-      @courses = Course.where(learning_partner_id: current_user.learning_partner_id)
+      @courses = Course.courses_only.where(learning_partner_id: current_user.learning_partner_id)
                        .where.not(neo_ai_course_id: nil)
                        .includes(:tags, banner_attachment: :blob)
                        .order(created_at: :desc)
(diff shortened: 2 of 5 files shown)


app/models/course.rb
```diff
@@ -8,6 +8,9 @@ class Course < ApplicationRecord
   PER_PAGE_LIMIT = 12
 
   enum :visibility, { public: 'public', private: 'private' }, prefix: :visibility
+  enum :kind, { course: 'course', microlesson: 'microlesson' }, prefix: :kind
+
+  scope :courses_only, -> { where(kind: :course) }
```
dr-r3d, line 13: `courses_only` exists on the model but is never called in `FilterService#base_course_scope`. Once microlessons exist in the table, `Course.all` and the `Course.where.not(...)` paths will include them in explore, home page, search, and enrolment listings without any further changes needed. The scope needs to be applied inside `base_course_scope` so existing listings are protected automatically.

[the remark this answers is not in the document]
sadrisya: Applied courses_only to all 5 entry points: the 4 branches in FilterService#base_course_scope and the non-admin path in CoursesController#manage

sadrisya: This is  fixed in commit 322bf9bb

PR #1458, opened 2026-07-01
Add microlesson wizard — Prompt and Configure Video steps (#1444)

### LINKED ISSUE #1444: Content Studio: microlesson wizard — Prompt & Configure steps
Implement the first two steps of the microlesson creation wizard.

## Steps
1. **Prompt** —  write a text prompt
2. **Configure Video** — choose avatar/template (populated from `list-templates`), optional logo upload

## On submit
- Call `ApiClient.create_microlesson` with the collected inputs
- Redirect to the Planning state screen (next issue)

## Figma
https://www.figma.com/design/QOquDODFlEr3jKFmyyw96z/AI-Course-Generation?node-id=3313-4137&m=dev

## Notes
- Depends on: Content Studio MockClient + ApiClient issue (#1443)
- Use NeoComponent helpers for all UI elements

sadrisya:
Closes #1444

## Summary
Adds the Microlesson creation wizard's first two steps — Prompt and Configure Video — with routing, controller, views, and Stimulus behavior for field persistence and validation.

## Related Issue
Closes [#1444](https://github.com/openvitae-tech/blackboard-lms/issues/1444)

## Changes

- Add routes: new, create, configure, update_config for microlessons/wizard
- Add WizardController handling prompt input and video config steps
- Prompt step: title field + textarea with 500-char counter (char_counter Stimulus controller)
- Configure step: template carousel + logo upload + background style radio (Image/Video/Plain)
- Add avatar_card helper to application_helper (for future use)
- Add char_counter_controller.js and upload_mode_controller.js Stimulus controllers
- Add microlesson.create_failed locale key
- Use the video-camera icon for the Configure Video step (was previously a generic camera icon)
- Persist the title/prompt fields in session across Back navigation from Configure → Prompt; only cleared on a fresh wizard start (?fresh=true) or after continuing to script review
- Disable "Continue to Video style" until both the title and prompt fields are filled (microlesson_prompt_controller.js)
- Bump text sizes on the Configure Video step (logo upload and background style copy) for readability
- 21 request specs covering both wizard steps, form submission, field persistence, and the fresh-start reset

## Screenshots / Visual Diffs
Screenshot 2026-07-02 at 5 30 50 PM
Screenshot 2026-07-02 at 5 37 36 PM

## Checklist


engines/content_studio/app/controllers/content_studio/microlessons/wizard_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      def configure
+        @microlesson_id = params[:id]
+        @templates = ApiClient.list_templates
+      end
+
+      def update_config
+        logo_url         = upload_logo(params[:logo])
+        template_id      = params[:template_id].presence
+        background_style = params[:background_style].presence
+        prompt           = session.delete(:ml_wizard_prompt)
```
dr-r3d: Session is cleared unconditionally before `ApiClient.create_microlesson` is called. If the API raises, the rescue redirects to `new_microlesson_path` with both `:ml_wizard_title` and `:ml_wizard_prompt` already gone — the user must retype everything.

Move the session teardown to after the successful redirect:

```ruby
def update_config
  logo_url    = upload_logo(params[:logo])
  template_id = params[:template_id].presence
  prompt      = session[:ml_wizard_prompt]   # read without deleting

  microlesson_id = ApiClient.create_microlesson(prompt: prompt, ...)

  clear_ml_wizard_session                    # only reached on success
  redirect_to configure_microlesson_path(id: microlesson_id, state: 'planning')
rescue StandardError => e
  redirect_to new_microlesson_path, alert: ...
  # session still intact — form re-populates
end
```

engines/content_studio/app/controllers/content_studio/microlessons/wizard_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        clear_ml_wizard_session
+
+        microlesson_id = ApiClient.create_microlesson(
+          prompt: prompt,
+          document_urls: [],
+          template_id: template_id,
+          logo_url: logo_url
```
dr-r3d: `background_style` is captured from params and logged but never forwarded to `ApiClient.create_microlesson`. The user's Image/Video/Plain selection will be silently discarded when the API method is added in the follow-up PR.

engines/content_studio/app/javascript/content_studio/controllers/char_counter_controller.js
```diff
@@ -0,0 +1,15 @@
...
+// Updates a character count display as the user types in a textarea.
+export default class extends Controller {
+  static targets = ['input', 'count']
+
+  connect() {
+    this.update()
+  }
+
+  update() {
+    const len = this.hasInputTarget ? this.inputTarget.value.length : 0
```
dr-r3d, line 12: No element in `new.html.erb` carries `data-char-counter-target="count"`, so `hasCountTarget` is always false and this line never runs. The "0 / 500" display is a static `support_text` string — the count never updates regardless of how much the user types.

You'll need to expose a live `count` target instead of hardcoding the zero in the support text.

engines/content_studio/app/controllers/content_studio/microlessons/wizard_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def create
+        session[:ml_wizard_title]  = params[:title].presence
+        session[:ml_wizard_prompt] = params[:prompt].presence
+        redirect_to configure_microlesson_path(id: :pending)
+      end
+
+      def configure
+        @microlesson_id = params[:id]
+        @templates = ApiClient.list_templates
```
dr-r3d, line 22: No rescue here — any `Faraday::Error` (network timeout, non-2xx from the templates endpoint) propagates unhandled and returns a 500. `update_config` in the same controller rescues `StandardError`; the omission in `configure` is inconsistent. `ApplicationController` has no `rescue_from` that would catch this.

```ruby
def configure
  @microlesson_id = params[:id]
  @templates = ApiClient.list_templates
rescue StandardError
  @templates = []
  flash.now[:alert] = I18n.t('content_studio.microlesson.templates_unavailable')
end
```

engines/content_studio/app/views/content_studio/microlessons/wizard/configure.html.erb
```diff
@@ -0,0 +1,112 @@
...
+          <p class="general-text-sm-normal text-letter-color-light">How your presenter's environment will look.</p>
+        </div>
+
+        <div class="flex flex-col gap-5">
+          <% [
+            { value: 'image', label: 'Image', subtext: '(A static scene)' },
+            { value: 'video', label: 'Video', subtext: '(Motion & animated)' },
+            { value: 'plain', label: 'Plain', subtext: '(Solid colour only)' }
+          ].each do |opt| %>
+            <label class="bg-white border border-line-colour rounded-lg p-3 flex items-center justify-between cursor-pointer">
```
dr-r3d: NeoComponent already exposes `input_radio` and `input_radio_component` helpers that cover this pattern. CLAUDE.md: *"Do not write raw HTML for anything a component already covers."* The raw `<label>` + `form.radio_button` approach skips NeoComponent's checked-state styling and won't pick up future token updates.

engines/content_studio/spec/requests/content_studio/microlessons/wizard_spec.rb
```diff
@@ -0,0 +1,126 @@
+# frozen_string_literal: true
```
dr-r3d, line 1: View specs are missing for the two new views. The courses wizard has `spec/views/content_studio/courses/wizard/new.html.erb_spec.rb` and `configure_video.html.erb_spec.rb` alongside its request spec — same pattern is expected here.

Needs:
- `spec/views/content_studio/microlessons/wizard/new.html.erb_spec.rb`
- `spec/views/content_studio/microlessons/wizard/configure.html.erb_spec.rb`

Also, `PATCH microlessons/:id/configure` (`update_config`) has no coverage of any kind.

engines/content_studio/app/javascript/content_studio/controllers/microlesson_prompt_controller.js
```diff
@@ -0,0 +1,25 @@
...
+// title and prompt fields are filled in.
+export default class extends Controller {
+  static targets = ['title', 'prompt', 'submit']
+
+  connect() {
+    this.updateSubmit()
+  }
+
+  updateSubmit() {
+    const hasTitle = this.hasTitleTarget ? this.titleTarget.value.trim().length > 0 : true
```
dr-r3d: When a target is absent the fallback is `true` (field treated as satisfied), which inverts the intent of a required-field guard. If a future template wires this controller without one of the inputs the button stays enabled unconditionally.

```js
const hasTitle  = this.hasTitleTarget  ? this.titleTarget.value.trim().length  > 0 : false
const hasPrompt = this.hasPromptTarget ? this.promptTarget.value.trim().length > 0 : false
```

engines/content_studio/app/controllers/content_studio/microlessons/wizard_controller.rb
sadrisya: session.delete(:ml_wizard_prompt) changed to session[:ml_wizard_prompt] so session is only read. clear_ml_wizard_session moved to after the successful API call so on failure the session stays intact and the form re-populates. Also passed bg_type: map_bg_type(background_style) to the API call so the user's background style selection is no longer silently discarded. Added rescue StandardError to configure so a list_templates network failure returns an empty array instead of a 500.

engines/content_studio/app/controllers/content_studio/microlessons/wizard_controller.rb
sadrisya: added bg_type: map_bg_type(background_style) to the create_microlesson call. Also added map_bg_type private method which translates the form value "plain" to "none" since Neo AI only accepts "video", "image", or "none".

engines/content_studio/app/javascript/content_studio/controllers/char_counter_controller.js
sadrisya: removed the static support_text string from textarea_component (which HTML-escapes its content so a <span> tag can't be injected). Added a <p> tag directly below the component containing <span data-char-counter-target="count">0</span> so the Stimulus controller can find the target and update the count live as the user types.

engines/content_studio/app/controllers/content_studio/microlessons/wizard_controller.rb
sadrisya: added rescue StandardError to configure with @templates = [] fallback and flash.now[:alert] using the new content_studio.microlesson.templates_unavailable i18n key, consistent with how update_config handles errors.

engines/content_studio/app/views/content_studio/microlessons/wizard/configure.html.erb
sadrisya: replaced raw <label> + form.radio_button with input_radio_component as per CLAUDE.md rule. The subtext was merged into the label string (e.g. "Image (A static scene)") since the component handles label rendering internally. This ensures checked-state styling and future design token updates are applied automatically.

engines/content_studio/spec/requests/content_studio/microlessons/wizard_spec.rb
sadrisya: added spec/views/content_studio/microlessons/wizard/new.html.erb_spec.rb and configure.html.erb_spec.rb following the same pattern as the courses wizard view specs. new spec covers wizard steps, field labels, char counter target, footer buttons, and session pre-fill. configure spec covers wizard steps, template/logo/background sections, all three radio options, empty-templates fallback, and footer buttons. update_config coverage was added in the request spec in the previous commit.

engines/content_studio/app/javascript/content_studio/controllers/microlesson_prompt_controller.js
sadrisya: changed both fallbacks from true to false. When a target is absent the field is now treated as unsatisfied, keeping the button disabled rather than incorrectly enabling it if a future template wires the controller without one of the inputs.

